PR #83, opened 2026-02-02
Issue74 model framework

julialuo02:


annawoodard (approved):
After 8ce82fb, a few review items from @jpivarski were still open on the branch. The main remaining issues were that ml_pipeline/pcr_prediction.py still had a broken vascular feature path (case_id was being coerced to float), use_radiomics was still a stub rather than a real runtime path, the clinical config key still did not match what  feature_factor.py expected, the runtime config and README were still out of sync with the current ml_pipeline/ layout, and the root `.gitignore` still had overly broad artifact rules. I had my agent address those remaining items in f90be28 so the branch is in a mergeable state, but please read that commit carefully and make sure you understand what was still missing after 8ce82fb and how it was fixed before we merge, since that is part of owning the PR. Once you’ve reviewed it and are comfortable explaining the fixes, please merge (you don't need to iterate again with us). (It's also possible some code just wasn't properly added to the PR)

PR #107, opened 2026-03-09
Vessel visualization

julialuo02:


julialuo02:
This is redundant code from issue74-model-framework

PR #125, opened 2026-04-06
Fix slurm sbatch wrappers to cd into REPO_ROOT before running python modules

Nedak23:
Fixes:
- Adds `cd "${REPO_ROOT}"` to the `CACHE_WRAP` and `MERGE_WRAP` sbatch commands in `submit_independent_signal_matrix_array.sh`

Why:
- Without this, `python -m` module resolution fails when the job's working directory is not the repo root (e.g. `ModuleNotFoundError`)


annawoodard (approved):
Awesome, thank you @Nedak23 👏 👏 👏

Issue #116, opened 2026-04-02
Extend the tabular pipeline and run a model-family ablation matrix


annawoodard:
<html><head></head><body><h2>Background</h2>
<p>The tabular feature pipeline already exists. This subproject is <strong>not</strong> about inventing new features — it's about comparing <strong>model families</strong> fairly while keeping the feature set fixed. Otherwise you can't tell whether a gain came from features or from the model.</p>
<p>The repo already has <strong>feature-arm ablations</strong> (which feature blocks carry signal?). This issue adds <strong>model-family ablations</strong> (which model family makes the best use of those blocks?).</p>
<h3>What already exists</h3>

File | Role
-- | --
train_tabular.py | Main tabular training entrypoint
tabular_models.py | Current sklearn model builder (logistic regression, random forest)
evaluation/ | Shared fold generation, metrics, saved outputs
configs/ispy2.yaml | Standard tabular run config
configs/independent_signal.yaml | Focused comparison against the clinical + tumor_size baseline

<p>The pipeline already supports numeric feature selection, nested tuning for logistic regression, stratified CV, site-exclusive group-aware CV, and per-stratum metrics.</p>
<h2>Key Definitions</h2>
<p><strong>Model selection</strong> — Try a small number of reasonable setups, compare them fairly, and choose based on training-data rules — not by hand-picking the best validation result after the fact.</p>
<p><strong>Nested tuning</strong> — Outer folds estimate real performance; inner folds choose hyperparameters. Safer than tuning and reporting on the same fold.</p>
<p><strong>Robustness</strong> — The result still holds under reasonable changes: site-exclusive folds, unilateral-only cohort, performance broken out by tumor subtype.</p>
<p><strong>Baseline arm</strong> — A simpler reference model (here, <code>clinical + tumor_size</code>) so you can ask "how much better is this than something simple?"</p>
<p><strong>Model-family ablation</strong> — Fix the feature arm, swap the model family, compare. Example: run the same <code>clinical + tumor_size + graph + kinematic</code> arm with logistic regression, random forest, and XGBoost.</p>
<h2>What To Do</h2>
<h3>1. Orient yourself</h3>
<p>Read the current tabular code and write a short summary of:</p>
<ul>
<li>which model families already exist</li>
<li>which feature-arm configs already exist</li>
<li>what result summaries are already tracked</li>
</ul>
<h3>2. Add at least one new model family</h3>
[...]

Issue #117, opened 2026-04-02
Run robustness analysis for the top tabular model families


annawoodard:
**Why this issue exists**

A single overall AUC is not enough. A model may look good under one split strategy but weaken under site-exclusive splits or for specific tumor subtypes.

**Main goal**

Take the top 1-2 model families from #116  and check whether they stay reasonably strong under a few important robustness tests.

**What to test**

At minimum:

- ordinary stratified cross-validation
- site-exclusive group-aware cross-validation
- performance by tumor subtype

Good optional checks:

- unilateral-only I-SPY2 subset
- comparison with and without site-related clinical columns

**Suggested procedure**

1. Freeze the top 1-2 model-family setups from #116.
2. Run those setups under:
   - standard CV
   - `use_group_split: true` with `group_col: "site"`
3. Keep `stratum_col: "tumor_subtype"` so the evaluator writes subtype metrics.
4. Compare:
   - overall AUC mean/std
   - fold variability
   - subtype-specific metrics

**What "robust" means here**

It does not mean the numbers are identical.

It means:

- performance does not collapse under site-exclusive folds
- no major subgroup is clearly much worse than the rest without explanation
- conclusions stay roughly the same across reasonable evaluation choices

**Definition of done**

- there is a summary table for standard CV vs site-exclusive CV for the top model families
- there is a short subgroup summary by tumor subtype
- there is a plain-English note saying which model family looks most robust

**Helpful file pointers**

- `evaluation/build_splits.py`
- `evaluation/README.md`
- `evaluation/evaluator.py`
- `configs/ispy2.yaml`

PR #126, opened 2026-04-07
Project 2: model-family x feature-arm ablation matrix (Issue #116 + 117)

danielgong1:
## Summary
- Adds model-family ablation support for tabular pCR modeling over fixed feature arms.
- Extends tabular models to include XGBoost (`lr`, `rf`, `xgb`).
- Adds reproducible runner/config/Slurm script plus a compact results snapshot.
## Key results (AUC mean)
- `clinical_plus_tumor_size`: `lr` **0.571** > `rf` 0.545 > `xgb` 0.527
- `clinical_plus_tumor_size_plus_kinematic`: `lr` **0.587** > `rf` 0.576 > `xgb` 0.562
- `clinical_plus_tumor_size_plus_vessel_all`: `xgb` **0.606** > `rf` 0.603 > `lr` 0.591
## Decision for next phase
- Keep **`lr` + `xgb`**.
- `lr` is strongest on simpler arms; `xgb` is best on the richest vessel arm.
## Repro
`cd slurm && CONFIG=../configs/model_family_matrix.yaml OUT_ROOT=../experiments/model_family_matrix_ispy2 sbatch submit_model_family_matrix.slurm`
Refs #116


danielgong1:
**#117 — Robustness:** Compared `lr` vs `xgb` on the full vessel arm under **standard CV** vs **site-exclusive CV** (`group_col: site`), with subtype metrics.
**Artifacts:** `run_model_family_robustness.py`, `configs/model_family_robustness.yaml`, `slurm/submit_model_family_robustness.slurm` · snapshots: `results/model_family_robustness_ispy2_summary.csv` + `..._subtype_summary.csv` · note: `docs/model_family_robustness_117.md`.
**TL;DR:** LR mean AUC drops under site splits; XGB mean AUC stays similar but more variable across folds. Refs #117

annawoodard (changes requested):
This is a great start! The experiment goal makes sense, but I don’t think the current amount of new orchestration code is justified.

Main request: Please rework this to reuse the existing ablation pipeline instead of adding two near-copy runner scripts.

  Right now `run_model_family_matrix.py` and `run_model_family_robustness.py` duplicate a lot of what `run_ablation_matrix.py` already does:
  - build one full feature table
  - slice per arm
  - write per-run configs/features
  - call `run_evaluation_pipeline`
  - summarize `metrics.json`

  I’d like this consolidated before merge, either by:
  - extending `run_ablation_matrix.py` to support `model_families` and optional split modes, or
  - extracting the shared matrix logic into a helper and keeping any new entrypoints very thin

  A few concrete issues also need to be fixed:

  1. `xgb` reproducibility
  `tabular_models.py` adds `XGBClassifier`, and the new configs enable `xgb` by default, but that only works if the project env is updated and documented. Please make sure the env used by the Slurm scripts actually has `xgboost` available, or gate/remove `xgb` until that is true.

  2. Nested tuning outputs get overwritten
  Both new runners call `run_evaluation_pipeline(..., runs_root)`, but `train_tabular.py` writes nested tuning artifacts directly into the passed `outdir`. That means nested-tuning summaries from different runs overwrite each other. Please fix this so each run writes its own tuning outputs.

  3. AP summary fields are inconsistent / not actually populated
  `run_model_family_matrix.py` looks for `"ap"`, while `run_model_family_robustness.py` looks for `"average_precision"`, and the committed robustness CSV already has blank AP columns. Please either implement AP properly in the evaluator and summarize it consistently, or remove those columns for now.

  4. New Slurm wrappers are fragile
  The new Slurm scripts use `../...` paths and don’t `cd` to repo root like the existing submission scripts do. Please align them with the existing Slurm pattern so they are robust to where `sbatch` is called from.

  I’m fine with the underlying modeling addition (`xgb` support in `tabular_models.py`), but I’d like the orchestration code simplified before merge. Please also make sure to report standard deviations over folds.

danielgong1:
@annawoodard — Thanks for the review. I consolidated orchestration into run_ablation_matrix.py, fixed per-run nested tuning outputs, aligned AP under ap in evaluation/metrics.py, added xgboost to environment.yml, and tightened Slurm (SLURM_SUBMIT_DIR → repo root, logs under logs/). Summary tables report AUC mean and AUC std over folds (auc_mean, auc_std).

Issue #116 — model-family matrix (5-fold CV), from results/model_family_matrix_ispy2_summary.csv:

clinical + tumor_size — LR 0.571 ± 0.046; RF 0.545 ± 0.060; XGB 0.527 ± 0.045
clinical + tumor_size + kinematic — LR 0.587 ± 0.041; RF 0.576 ± 0.038; XGB 0.562 ± 0.059
clinical + tumor_size + morph + graph + kinematic — LR 0.591 ± 0.034; RF 0.603 ± 0.038; XGB 0.606 ± 0.028
Issue #117 — robustness on full vessel arm, from results/model_family_robustness_ispy2_summary.csv:

LR — standard CV 0.595 ± 0.032; site-exclusive CV 0.523 ± 0.060
XGB — standard CV 0.606 ± 0.028; site-exclusive CV 0.608 ± 0.054

annawoodard (changes requested):
Thanks @danielgong1, this is great!

The consolidation into `run_ablation_matrix.py`, the per-run nested tuning fix, the AP alignment, and the Slurm cleanup address my main code-structure concerns.

The remaining thing I would want for this to feel complete is just a bit more result framing:
  - For `#116`, please add one short conclusion (comment is fine) that states exactly which families you recommend carrying forward and why, based on the matrix.
  - For `#117`, please add one short conclusion that states exactly what robustness claim you think is supported by the site-exclusive result.

  The numbers themselves are enough for me; I mainly want the final takeaway written down somewhere

danielgong1:
Changes implemented!

annawoodard (approved):
Awesome!!! 👏 👏 👏 👏 

Thanks for your great work on this @danielgong1

Issue #122, opened 2026-04-02
Add configurable pooling variants for the Deep Sets model


annawoodard:
# Deep Sets Modeling and Evaluation Subproject

## Process Note

- This subproject should mostly treat the point features from the Point-Feature Engineering Subproject as fixed. The goal is to improve the model and evaluation path, not to keep changing the point-set builder.
- Training runs should go through Slurm and the `vanguard` environment.

## Current Baseline

The Deep Sets training path already exists as a starter baseline.

**Model architecture:**

- Encode each point with an MLP
- Pool encoded points with mean, max, and log point count
- Classify the pooled case representation with another MLP

**Training loop:**

- Adam optimizer
- Weighted `BCEWithLogitsLoss`
- Fixed number of epochs (no early stopping, no best-checkpoint selection)
- No configurable pooling or loss choice

**Existing outputs:**

- Evaluator metrics and plots
- Fold prediction tables
- Loss history by epoch
- Probability distribution plot and summary JSON

## Key Definitions

**Pooling** — How the model turns many point embeddings into one case embedding (e.g., average all points, take the max, or combine both).

**Loss / Objective** — The training score the model tries to minimize (e.g., binary cross-entropy for pCR vs. non-pCR).

**Training stabilization** — Changes that make learning less noisy or brittle (e.g., early stopping, gradient clipping, learning-rate scheduling, best-epoch checkpoint selection).

**Diagnostics** — Saved numbers and plots that help us understand what happened during training (e.g., loss curves, predicted probability distributions, subgroup metrics, point-count summaries).

---

### Why this issue exists

The current model uses one fixed pooling recipe. That is a reasonable baseline, but it is not enough to answer whether the learned set model is limited by its pooling step.

### Goal

Make the Deep Sets model support a small number of pooling variants and compare them fairly.

### Pooling variants to start with

- mean only
- max only
- sum only
- mean + max
- mean + max + log_count (close to the current baseline)

### Steps

1. Read the current model and trainer; write a short summary of the current point features, pooling behavior, loss/optimizer setup, and saved outputs.
2. Add a config-controlled pooling choice.
3. Keep the rest of the model fixed.
4. Compare the pooling variants on the same point-feature dataset and fold setup.
5. Save one compact result table.

### Advice

Don't add too many exotic pooling ideas at once. A good first question is simply: does the model work better when it remembers average behavior, extreme behavior, or both?

### Definition of done

- Pooling choice is configurable
[...]

PR #127, opened 2026-04-07
Add Configurable pooling variants for the Deep Sets model

Nedak23:
Addresses #122

- Makes DeepSetsClassifier pooling configurable via a pooling config key, supporting: mean, max, sum, mean_max, mean_max_logcount (default)
- Adds the option to DEFAULT_CONFIG and configs/deepsets_ispy2.yaml with backward-compatible defaults
- Fixes max-pooling to use a portable loop (compatible with PyTorch <1.12)
Adds 22 parametrized tests in tests/test_deepsets_model.py


jpivarski (approved):
This addresses some points in #122, but not all of them, so I removed it from the "would close" list. The issue also asks for:
  - a short summary of the current baseline setup,
  - fair comparison of pooling variants on the same dataset/folds,
  - one compact result table,
  - selection of a default pooling based on that comparison.

But this PR does its focused task: it implements configurable pooling in the model/training path, and it doesn't seem to have any bugs. The test coverage checks shapes, gradients, defaults, and invalid names, but it doesn't check to see if the pooling is correctly implemented with any numerical tests.

It's in a good state to merge before moving on to the next PR.

Issue #119, opened 2026-04-02
Audit the current Deep Sets point-set builder and propose stronger point features


annawoodard:
## Background
 
The Deep Sets path already exists, but it's intentionally minimal. Right now the builder keeps tumor-local skeleton points and stores only one point feature: `curvature_rad`. This subproject is about building **better point sets and better point features** — not changing the model architecture.
 
### What already exists
 
| File | Role |
|---|---|
| `build_deepsets_dataset.py` | Builds and saves one point set per case |
| `deepsets_data.py` | Data loading for training |
| `train_deepsets.py` | Deep Sets training entrypoint |
| `graph_extraction/README.md` | Documents the upstream centerline/graph outputs |
 
**Current inclusion rule:** keep skeleton points whose signed distance to the tumor is ≤ the local vessel radius. If nothing passes, fall back to the nearest 64 points.
 
**Current point features:** just `curvature_rad`.
 
**Full dataset builds should run through Slurm, not on the headnode.**
 
## Key Definitions
 
**Point** — One vessel/skeleton voxel included in a case's point set.
 
**Point feature** — A number attached to each point (e.g. curvature, vessel radius, signed distance to tumor).
 
**Inclusion rule** — The filter that decides which points make the cut (e.g. all points within 50 mm of the tumor, or the nearest N points).
 
**Manifest** — The case-level CSV that tells training where each saved point set lives.
 
### 1. Audit the current builder
 
Read `build_deepsets_dataset.py`, `deepsets_data.py`, and `train_deepsets.py`. Write a short summary of:
 
- What is currently stored per point
- What is stored per case in the manifest
- What metadata are available from the upstream centerline/graph outputs but currently dropped
 
### 2. Propose candidate point features
 
Make a plain list. For each candidate, note:
 
- **Feature name**
- **Short description**
- **Source files needed**
- **Whether it can be computed from existing saved outputs**
 
Here are strong candidates to consider:
 
- Signed distance to tumor boundary
- Absolute distance to tumor boundary
- Inside-tumor flag
- Shell bin or shell one-hot encoding
- Normalized XYZ offset from tumor centroid
- Local vessel radius from the support mask
- Node degree
- Endpoint / chain / bifurcation flags
- Local direction relative to the tumor centroid
- Point-local dynamic features sampled from the DCE time series
- Point-local graph-context features read from saved graph outputs
 
### 3. For dynamic point features, verify spatial alignment
 
[...]

PR #129, opened 2026-04-07
Fix/deepsets slurm out root and nounset

danielluna90:
Deepset example did not work initially, following the procedure listed in the README.md with a clean environment. Took some time to debug it, this is the product. Going to link this to Issue #119.


slurm/deepsets_job.slurm
```diff
@@ -7,7 +7,7 @@
 #SBATCH --output=logs/deepsets-%x-%j.out
 #SBATCH --error=logs/deepsets-%x-%j.err
 
-set -euo pipefail
+set -eo pipefail
```
danielluna90, line 10: This seems to make micromamba happy. When starting the SLURM job it raises a variable unset error.

jpivarski (changes requested):
Issue #119 is asking for the whole feature engineering. This PR seems to be just to get the minimal example running.

Within that scope, it looks okay apart from one issue:

The new PyTorch-compat fallback in [deepsets_model.py:106](/home/jpivarski/dsi/vanguard/deepsets_model.py#L106) is broken. `torch.scatter_reduce` is called as `torch.scatter_reduce(encoded, 0, idx, reduce="amax", output_size=batch_size)`, but the actual signature is `torch.scatter_reduce(input, dim, index, src, reduce, *, include_self=True)`. It requires a `src` argument and does not accept `output_size`. On a runtime that lacks `Tensor.scatter_reduce_` and needs this fallback, the model will fail at inference/training start with a `TypeError`. That means the claimed compatibility fix does not actually work for the old-PyTorch case it targets.

jpivarski (approved):
The problem above has been resolved (checked by inspecting the code, not attempting to run in an older PyTorch to prove compatibility).

This PR addresses a smaller problem than issue #119, so it's important that merging this does _not_ close the issue.

Auto-merge is on, so as soon as I accept the changes, it will merge. Here it goes!

Issue #123, opened 2026-04-02
Compare a small set of loss and objective choices for Deep Sets


annawoodard:
**Why this issue exists**

The current Deep Sets trainer always uses one weighted binary cross-entropy setup. That is a fine baseline, but we should check whether a small number of alternative objectives behave better on this class-imbalanced problem.

**Main goal**

Add a small, manageable set of objective choices and compare them fairly.

**Recommended starting set**

Start with:

- current weighted `BCEWithLogitsLoss`
- unweighted `BCEWithLogitsLoss` as a reference
- focal loss as the main alternative

Optional stretch:

- label smoothing
- threshold tuning on training folds after probability prediction

**Student-friendly advice**

Keep this question simple:

- does a different training objective make validation performance better or more stable?

If a new loss only changes the number a tiny amount but makes training much harder to explain, that may not be worth it.

**Implementation guidance**

- add config keys for objective choice
- keep the optimizer and architecture fixed at first
- compare losses on the same point-feature dataset and same folds

**Definition of done**

- loss choice is configurable
- at least one alternative loss beyond the baseline is implemented
- results are summarized in one small table
- one default loss choice is selected

**Helpful file pointers**

- `train_deepsets.py`
- `deepsets_model.py`
- `configs/deepsets_ispy2.yaml`

Nedak23:
## Loss Function Comparison Results

Ran all 3 loss variants on the ISPY2 cohort

| Loss | Mean AUC | Std AUC |
|---|---|---|
| weighted_bce | 0.532 | 0.029 |
| unweighted_bce | 0.541 | 0.026 |
| **focal** | **0.546** | **0.020** |

<img width="1600" height="800" alt="Image" src="https://github.com/user-attachments/assets/fd556f26-bdd9-4187-8aba-7c733c46cab7" />

Focal loss has the highest mean AUC and the lowest variance. The differences are modest (~1.4 AUC points between focal and weighted BCE), but focal is consistently more stable across folds.

**Selected default: `focal`** — marginally better mean AUC and noticeably tighter fold variance. Updated `config.py` default

PR #133, opened 2026-04-13
Add configurable loss function for Deep Sets training (#123)

Nedak23:
Closes the code-infrastructure portion of #123. Does **not** include the experimental comparison or results table

## What this PR covers
- Adds a `loss` config key (`weighted_bce`, `unweighted_bce`, `focal`) so the Deep Sets training objective is selectable via YAML config
- Implements `FocalWithLogitsLoss` and a `build_loss_fn` factory that replaces the previously hardcoded weighted BCE loss
- Adds `focal_alpha` and `focal_gamma` config keys for tuning focal loss
- Default remains `weighted_bce`, so existing runs are unchanged

## What still needs to happen (per #123)- waiting on better features/toy feature
- Run experiments comparing the three losses on the same folds
- Summarize results in a comparison table
- Select a default loss choice based on the results


jpivarski (approved):
No errors observed when I tested it and the code implements what the PR says it does (not the whole of issue #123).

PR #136, opened 2026-04-14
Add training stabilization, metadata, and diagnostics (#124)

### LINKED ISSUE #124: Improve Deep Sets training stabilization, metadata-aware evaluation, and diagnostics
**Why this issue exists**

Even a reasonable model can be hard to trust if training is noisy and the outputs are hard to interpret. The current trainer saves useful basics, but it is missing several things that make experiments easier to compare and debug.

**Main goal**

Make the Deep Sets training path more stable and easier to analyze.

**High-value improvements**

Training stabilization:

- early stopping
- save or restore the best validation epoch
- gradient clipping
- optional learning-rate scheduler
- regularization, for example dropout and/or weight decay

Evaluation and metadata:

- carry useful metadata into the Deep Sets manifest:
  - `site`
  - `tumor_subtype`
  - `bilateral`
- use that metadata for:
  - per-subtype metrics
  - optional site-exclusive group-aware splits

Diagnostics:

- save best epoch per fold
- save point-count summaries by fold
- save fallback-case summaries
- flag obviously collapsed probability behavior

**Why the metadata part matters**

`train_deepsets.py` is already ready to use subgroup columns if they exist in the manifest, but the current Deep Sets manifest writer mostly drops them.

So one useful part of this issue is:

- write those fields into the manifest during dataset build
- make sure training/evaluation reuses them

**Advice**

This issue is not about adding everything possible.

A very strong result here would be:

- training is less noisy
- the chosen epoch is more defensible
- subgroup metrics are easier to inspect
- the final run outputs are easier for the whole team to read

**Definition of done**

- at least one stabilization or regularization method is implemented
- best-epoch or early-stopping behavior is saved clearly
- Deep Sets manifest includes useful subgroup metadata
- Deep Sets results can report subgroup metrics when available
- there is a short note explaining which diagnostics are now saved

**Helpful file pointers**

- `train_deepsets.py`
- `build_deepsets_dataset.py`
- `merge_deepsets_manifest.py`
- `evaluation/evaluator.py`
- `evaluation/build_splits.py`

Nedak23:
## Summary

- **Training stabilization:** early stopping with best-epoch model restoration, gradient clipping, and optional LR scheduler (cosine or plateau). All controlled via config and disabled by default.
- **Metadata propagation:** writes `site`, `tumor_subtype`, and `bilateral` into the Deep Sets manifest, unlocking the already-implemented subgroup metrics and group-aware splits.
- **Fold diagnostics:** saves `fold_diagnostics.csv` with best epoch, point-count stats, fallback case counts, and collapsed-probability detection per fold.
- **Documentation:** expanded `run_deepsets_pipeline` docstring documenting all diagnostic outputs and stabilization features.

Closes #124


build_deepsets_dataset.py
```diff
@@ -423,6 +423,9 @@ def main() -> None:
                 "set_path": str(set_path),
                 "label": int(row.label),
                 "dataset": dataset_name,
+                "site": str(getattr(row, "site", "")),
+                "tumor_subtype": str(getattr(row, "tumor_subtype", "")),
+                "bilateral": str(getattr(row, "bilateral", "")),
```
jpivarski: By converting each of these values into strings, a Pandas missing value would get converted into the string `"nan"`. Are you sure that `site`, `tumor_subtype`, and `bilateral` are never `"nan"` as distinct from missing?

jpivarski: In

https://github.com/dsi-clinic/vanguard/blob/a5d6fad083c079132545ac529e4315dc3b8932ed/evaluation/metrics.py#L172

the `dropna()` logic won't work for stringified `"nan"`.

jpivarski (approved):
It looks like this git branch implements the PR's scope and the PR's scope covers the issue's scope, so it's good!

I was wondering at what seemed like skipping regularization—we will almost certainly need it—but it looks like the model already had dropout.

One potential bug (see below). If you verify that it's not an issue, go ahead and merge. Otherwise, be sure to fix it!

train_deepsets.py
```diff
@@ -466,14 +491,91 @@ def fit_predict_one_fold(
...
+            logging.info(
+                "fold %d early stopping at epoch %d (best epoch %d, best val_loss %.4f)",
+                split.fold_idx,
+                epoch_idx + 1,
+                best_epoch,
+                best_val_loss,
+            )
+            break
+
+    if best_state_dict is not None:
```
Nedak23: This restores the lowest-validation-loss checkpoint unconditionally, even when `early_stopping_patience` is left at its default `0`. That means the new stabilization behavior is not actually disabled by default: a default run now predicts from the best epoch instead of the final epoch. The PR description says these features are disabled by default: https://github.com/dsi-clinic/vanguard/pull/136

Issue #120, opened 2026-04-02
Add stronger geometric, topology, and dynamic point features and benchmark the feature sets


annawoodard:
**Why this issue exists**

Right now each point has only one feature. That is too weak for a learned set model to do much beyond counting and averaging one signal.

This issue should also connect clearly to the tabular pipeline. The repo already has a strong `kinematic` tabular block that summarizes how vessel signal changes over time near the tumor. Deep Sets should try to carry some of that same dynamic information, but at the point level instead of only as case-level summary features.

**Main goal**

Upgrade the saved Deep Sets case sets so each point has a small, useful feature vector instead of just `curvature_rad`, including dynamic features that are similar in spirit to the existing tabular kinematic features.

Then run a simple benchmark over the existing Deep Sets baseline config so we can compare feature sets fairly.

**What "similar to the kinematic tabular features" means**

The tabular `kinematic` block already captures ideas like:

- when enhancement appears
- how strong enhancement is
- how quickly signal rises or falls
- how behavior changes by tumor-distance shell
- whether near-tumor vessels behave differently from farther-away vessels

For Deep Sets, we do not need to recreate the full tabular feature block exactly.

Instead, we want point-level or very local versions of the same ideas, such as:

- pointwise arrival / enhancement time
- pointwise peak signal or peak change
- pointwise wash-in slope
- pointwise wash-out slope
- pointwise total positive enhancement over time
- pointwise reference-normalized signal if there is a clean reference signal source
- shell membership so the model can learn near-vs-far dynamic differences

If graph-local dynamic features are feasible, good examples are:

- difference between a point's dynamic summary and the mean of its graph neighbors
- simple proximal-vs-distal contrast along the local centerline

If those are too hard for the first pass, it is completely fine to start with pointwise dynamic curve summaries only.

**Recommended feature-set groups**

Start by thinking in feature sets, not one feature at a time.

Suggested groups:

1. current baseline
   - `curvature_rad`
2. geometry + topology
   - `signed_distance_mm`
   - `abs_signed_distance_mm`
   - `inside_tumor`
   - shell flags such as `shell_inside`, `shell_0_2mm`, `shell_2_5mm`, `shell_5_10mm`
   - `degree`
   - `is_endpoint`
   - `is_chain`
   - `is_bifurcation`
   - tumor-centered offsets such as `offset_x_mm`, `offset_y_mm`, `offset_z_mm`
   - `support_radius_mm` if available
3. geometry + topology + dynamic
[...]

PR #137, opened 2026-04-15
Deep Sets: multi-regime point features (geometry, topology, kinetics) and benchmark configs

danielluna90:
Extends the Deep Sets dataset builder so each skeleton point can carry a stable, ordered feature vector instead of only curvature_rad, and adds pinned YAML configs for three benchmark arms: baseline curvature, geometry/topology, and geometry/topology plus vessel time-series summaries aligned with the tabular kinematic block.

Training hyperparameters are unchanged across arms; comparisons are by feature set only.


jpivarski (approved):
One quick thing: "draft PR" and "requesting review" are mutually exclusive. If it's a draft, then you're asking for it to not be merged, but if you're requesting review, you're asking if it is ready to be merged. I think you meant to take it out of draft, so I'll do that.

This PR is narrower than issue #120, so I detached it. THe issue requires not just builder/config changes, but also a benchmark table across the three feature sets and a short conclusion about whether they helped. You can do that in a PR comment—in fact, that's a good place to do it, so it remains attached to the PR and issue—but it hasn't been done.

I couldn't reproduce this because NumPy 2 doesn't have the `np.trapz` function for integration. OMG—this repo asks for NumPy 1:

https://github.com/dsi-clinic/vanguard/blob/1562c0ccebbcbc1014a19597f63a451157de22ee/environment.yml#L11

Okay, then I need to test it in the pinned environment. ...3 GB of package installations later: it works. This can be merged; I'll put a comment on #120 about what remains to be done.

Issue #138, opened 2026-04-15
QC Viz for tabular features


annawoodard:
• Add a quick QC visualization that, for each numeric feature, plots overlaid histograms for pcr=0 and pcr=1. The goal is not to prove predictive value, just to catch obviously garbage features: all-constant
  columns, extreme outliers, nearly empty columns, weird spikes, or distributions that look clearly broken. Save the plots in a simple reproducible folder so we can scan them before trusting the engineered table.

annawoodard:
bonus points: order it by the separation between pcr=1 and pcr=0

PR #140, opened 2026-04-15
Add 1st order and 2nd order visualization code 

RebeccaRuochunWu:
#138

image

image


annawoodard (approved):
This is a useful start on `#138`, but I would describe it as only partially addressing the issue.

  What it does add:
  - a reproducible QC plotting script
  - output files written to a chosen folder instead of committing generated figures
  - summary CSVs for the plotted features

  The main reason I say “partial” is that the script currently plots one pooled histogram per feature. It does not yet do the key comparison the issue asked for: overlaid `pcr=0` vs `pcr=1` histograms.

  Two smaller scope gaps:
  - it only plots the second-order features plus a hand-maintained list of first-order source columns, not “each numeric feature”
  - it clips to the 1st–99th percentile before plotting, which may hide some of the exact outliers or strange tails this QC pass is supposed to catch

  Also, this PR is broader than `#138`, since it includes the second-order feature-engineering changes as well, not just the visualization script.

  So overall: useful infrastructure and merging now as a waypoint is fine, but `#138` is not fully closed by this PR as written.

PR #142, opened 2026-04-21
QC Viz by PCR code

RebeccaRuochunWu:
solves issue #138 
viz results are the following:
image
image

Empty plots explanation

Second-order plot (distributions_second_order_by_pcr.png):

 - kinematic_crossing_early_fraction — The underlying source column (kinematic_early_enhancing_fraction_count_fraction_crossing) was
 never written to the per-study JSON payloads. The computation code exists in kinematic.py (line 688) but was added after the JSON
payloads were generated. Fixing this requires re-running the full kinematic extraction pipeline via Slurm across all 808 studies.

First-order plot (distributions_first_order_by_pcr.png):

 - tumor_size_tumor_voxels, tumor_size_tumor_shell_r0_2_voxels, tumor_size_tumor_shell_r4_8_voxels — These 3 columns are absent from
 the CSV because the second-order experiment was built without include_tumor_local=True (or the tumor mask root path wasn't found at
 build time). The data exists in features_full_raw.csv and was merged in for the second-order plot fix, but these first-order source
 columns remain missing from the first-order plot since they weren't in the original CSV's column set.
 - graph_skeleton_voxels, graph_nodes, graph_edges — Present in the CSV but entirely NaN. These are listed as first-order source
columns in the plotting script but don't appear to be produced by the current graph feature extraction pipeline.


annawoodard (approved):
Nice work, but I think this needs one fix before merge.

The new scripts are documented as running from the repo root with `python scripts/...`, but that invocation fails because Python adds `scripts/` to `sys.path`, so the repo root is not importable. I reproduced this for both scripts: `scripts/plot_pcr_distributions.py` and `scripts/add_second_order_features.py`.

  Both fail with:

  ```text
  ModuleNotFoundError: No module named 'features'
```

  A good fix would be to add the repo root to sys.path before importing from features, or otherwise change the documented invocation to require PYTHONPATH=.. I’d prefer that the script handles this itself, so the usage example works as written.

Approving to keep this moving, contingent on fixing the script import issue before merge. Sorry for the late review here.

Issue #130, opened 2026-04-09
Clean Up Redundant Tabular Features and Re-run Baseline Logistic Regression


RebeccaRuochunWu:
# Clean Up Redundant Tabular Features and Re-run Baseline Logistic Regression

  ## Why this issue exists

  We now have many engineered tabular features, including new second-order ones. Some of them are probably telling us almost the same thing.

  This can be a problem because:
  - the model may have many near-duplicate columns
  - results become harder to interpret
  - feature selection may choose different "versions of the same feature" in different folds

  The goal of this issue is to clean up highly redundant features, then check whether baseline logistic regression works just as well or better on the cleaner feature set.

  ## Main goal

  Find groups of highly correlated engineered features, keep the most useful version in each group, and compare baseline logistic regression before vs after pruning.

  ## What to do

  1. Build the full tabular feature table.
  2. Look for highly correlated numeric features.
     - Start with a simple threshold like `|r| >= 0.95`.
  3. Group together features that are basically duplicates.
  4. For each group, choose which feature to keep.
     - Prefer features that are:
       - easier to understand
       - more normalized
       - less missing
       - less repetitive
  5. Create a cleaned/pruned feature set.
  6. Run baseline logistic regression on:
     - the current feature set
     - the cleaned feature set
  7. Compare the results.

  ## What to compare

  At minimum:
  - AUC mean/std
  - number of features kept
  - whether the cleaned feature set is easier to explain

  ## Important note

  This issue is not mainly about trying lots of different models.

  Keep the model simple:
  - use baseline logistic regression

  The main question is:
  - does feature cleanup help us get a smaller, clearer feature set without hurting performance much?

  ## Definition of done

  - there is a table or note showing groups of highly correlated features
  - there is a clear keep/drop decision for those groups
  - baseline logistic regression has been rerun on the cleaned feature set
  - there is a short summary saying whether the cleaned set should become the new default

  ## Helpful file pointers

  - `tabular_cohort.py`
  - `tabular_models.py`
  - `train_tabular.py`
  - `features/second_order.py`
  - `features/README.md`
  - `configs/ispy2.yaml`

PR #143, opened 2026-04-21
Add 3-stage feature selection pipeline for pCR prediction

RebeccaRuochunWu:
resolves issue #130 
 **Pipeline**

  Three-stage feature selection with nested cross-validation to prevent data leakage:

 *Stage 1: Pre-filter*

  Drop NaN, near-constant, and correlated (|r|>0.8) features. Correlation tie-breaking is unsupervised (coverage + MAD).

  *Stage 2: Univariate screening — Mann-Whitney U with BH FDR correction*

  For each feature, a Mann-Whitney U test checks whether the distribution differs between pCR and non-pCR groups. This is a non-parametric rank-based test — it doesn't assume normality, which suits our imaging
  features.

  Testing hundreds of features creates a multiple testing problem: at p < 0.2, we'd expect ~60 false positives by chance alone. Benjamini-Hochberg (BH) FDR correction addresses this by adjusting p-values upward
  based on their rank, controlling the expected false discovery rate — i.e., the proportion of "discoveries" that are actually noise. After BH correction at α=0.2, only features with genuine signal survive.

  *Stage 3: Elastic-net logistic regression*

  Elastic net combines L1 (lasso) and L2 (ridge) regularization (l1_ratio=0.5):

   - L1 drives some coefficients exactly to zero → automatic feature selection
   - L2 shrinks correlated features together instead of arbitrarily picking one

  This is better than pure L1 for our data, where many features are highly correlated (e.g., the ~950 kinematic features across shells). Pure L1 would randomly pick one from each correlated group; elastic net
  retains the group signal.

  The regularization strength (alpha) is tuned via inner cross-validation.

  All three stages run inside each outer CV fold (5×3 RepeatedStratifiedKFold) to produce unbiased performance estimates.

  **Results**

  1,179 → 294 → 49 → 4 selected features

  Nested CV AUC: 0.591 ± 0.080 (unbiased estimate)
image

image
  

**Final 4 features**

Screenshot 2026-04-29 at 12 23 50 PM

image

  **Key changes from v1**

   - Fixed data leakage: stages 1–2 previously ran on all data before CV, inflating AUC
   - Added BH FDR correction: reduces false positives in screening (82 → 49 features)
   - Switched to elastic net: retains correlated feature groups better than pure L1
   - Removed misleading in-sample metrics: only nested CV AUC is reported
   - Added visualizations: per-fold ROC curves, AUC distribution, feature stability, raw vs adjusted p-values, coefficients


annawoodard (approved):
Thanks, Rebecca! Great work. I’m approving this so we can keep moving, but please make the changes below before we treat these results as final.

This PR is useful as a leakage-aware marginal/linear feature-selection baseline: the supervised screening and elastic-net fitting are inside the CV folds, which is exactly the right direction compared with doing feature selection globally before CV.

A few important fixes/clarifications:
1. Please fix the correlated-feature pruning logic in `stage1_prefilter`. Right now, if the feature `kept[i]` is marked for dropping, the loop can still keep comparing it to later columns and dropping additional features that may not be redundant with the retained representative. The safest fix is to form correlated feature groups/components and choose one representative per group using the unsupervised score, or at a minimum, stop comparing from `kept[i]` once it is marked for dropping.
2. Please flag when Stage 2 falls back to raw p-value top-k. I’m fine with a fallback so the pipeline doesn't fail, but the output should clearly state that the selected features did not pass BH/FDR correction. Otherwise, it is easy to describe the result as FDR-controlled when it was actually a raw-p fallback.
3. Please soften “definitive feature list.” The final full-data refit is useful for producing a candidate feature list, but because Stage 2 is univariate and the final coefficients are based on all data, I’d call this a “candidate marginal/linear feature list” rather than the definitive best feature set.

One scientific limitation to state explicitly: Mann-Whitney + BH/FDR is a marginal screen. It can miss features that only matter jointly or nonlinearly. That does not make this approach wrong; it just means this is the linear/marginal baseline, and we should follow it with a nonlinear comparator (you've put in #149).

Please address these and merge

PR #144, opened 2026-04-22
updating config.py default to focal

### LINKED ISSUE #123: Compare a small set of loss and objective choices for Deep Sets

Nedak23:
In light of testing results, updating config.py default to focal to complete Issue 123 for now.
loss_comparison

### DIFF
--- a/config.py
+++ b/config.py
@@ -138,7 +138,7 @@ DEFAULT_CONFIG: dict[str, Any] = {
         "lr_scheduler": "none",
         "lr_scheduler_factor": 0.5,
         "lr_scheduler_patience": 5,
-        "loss": "weighted_bce",
+        "loss": "focal",
         "focal_alpha": 0.25,
         "focal_gamma": 2.0,
         "deepsets_local_radius_floor_mm": 50.0,


jpivarski (approved):
I read your more detailed description of the results at https://github.com/dsi-clinic/vanguard/issues/123#issuecomment-4297333639.

Looks good! Well, the model quality is very marginal, but it's apparently the best that can be done.

This PR satisfies all the remaining parts of issue #123, so you can squash-and-merge whenever you're ready!

PR #147, opened 2026-04-29
Document issue #120 benchmark metrics

danielluna90:
Changes
- Fill issue-120 benchmark runbook table with concrete baseline and geom+topo AUC values from pinned-arm metrics artifacts.
- Add interpretation note comparing baseline vs geom+topo and explicitly flag dynamic-arm artifact gap.


jpivarski (approved):
As I see it, this PR doesn't resolve #120 because it just documents partial benchmark results for issue #120.

The issue asks for upgraded Deep Sets point features, stable feature_names, successful dataset build/merge, and a benchmark across feature sets. I detached #120 and this PR is mergeable as-is.

PR #153, opened 2026-05-05
Issue #119: Migrate Deep Sets audit/design note to executable notebook

### LINKED ISSUE #119: Audit the current Deep Sets point-set builder and propose stronger point features

danielluna90:
## Summary
- Replaces the static Issue #119 design note with `analysis/deepsets_issue119_notebook.ipynb`.
- Removes superseded markdown doc `docs/deepsets_issue119/README.md` and updates cross-links.
- Updates alignment tooling defaults so notebook-driven outputs go to analysis paths (`scripts/deepsets_alignment_check.py`).
 
## What changed
- Added: `analysis/deepsets_issue119_notebook.ipynb`
- Deleted: `docs/deepsets_issue119/README.md`
- Updated: `graph_extraction/README.md`
- Updated: `scripts/deepsets_alignment_check.py`


jpivarski (approved):
This PR completes the work of issue #119. Merge it! `:)`

PR #154, opened 2026-05-05
Issue #120: Add new Deep Sets point-feature regimes and migrate benchmark docs to notebook

### LINKED ISSUE #120: Add stronger geometric, topology, and dynamic point features and benchmark the feature sets

danielluna90:
## Summary
- Adds three new point-feature regimes using existing per-point signals:
  - geometry_topology_plus_curvature
  - geometry_topology_no_shells
  - curvature_plus_dynamic
- Adds three pinned configs for deferred runs (prepared now, execution later).
- Replaces Issue #120 markdown docs with executable notebook deliverable and generated analysis artifacts.

### Core implementation
- Updated build_deepsets_dataset.py:
  - expanded VALID_DEEPSETS_POINT_FEATURE_SETS
  - extended deepsets_point_feature_names(...)
  - added row assembly logic for 3 new regimes
  - preserved row/name consistency validation checks
- Updated `tests/test_deepsets_point_features.py` for new regimes and feature lengths
- Added configs:
  - configs/deepsets_ispy2_pointfeat_geom_topo_plus_curvature.yaml
  - configs/deepsets_ispy2_pointfeat_geom_topo_no_shells.yaml
  - configs/deepsets_ispy2_pointfeat_curvature_plus_dynamic.yaml

### Docs/notebook migration
- Added:
  - `analysis/deepsets_issue120_notebook.ipynb`
  - `analysis/figures/deepsets_issue120/feature_columns_by_config.csv`
  - `analysis/figures/deepsets_issue120/feature_set_auc_comparison.png`
  - `analysis/figures/deepsets_issue120/feature_set_benchmark_summary.csv`
- Deleted:
  - `docs/deepsets_issue120_benchmark.md`
  - `docs/deepsets_issue120_alignment.md`
- Updated links/indexes:
  - `analysis/README.md`
  - `docs/deepsets_issue119/README.md` (cross-link to Issue #120 notebook)


danielluna90:
This update finishes the Issue #120 Deep Sets point-feature benchmark with eight registered arms, an analysis notebook, and summary artifacts, with metrics for six of eight arms. The two missing dynamic ISPY2 arms still need a rerun with the updated Slurm defaults. The build → merge → train path now has stage timing, a shared per-case build cache, vectorized dynamic features, optional intra-shard parallelism, and skip/resume for completed stages. scripts/deepsets_sacct_summary.sh and the extended benchmark driver make reruns and elapsed-time tracking easier on Slurm.

jpivarski (approved):
It doesn't look like the missing benchmarks mentioned in https://github.com/dsi-clinic/vanguard/issues/120#issuecomment-4255564142 are done, but at this point, I think we can let that go.

The rest of this PR is other content: new features, configuration files, and Slurm tooling. Let's merge these things now, but keep in mind for your future projects that it's much better for coordination if a PR addresses the same scope as the issue it targets.

Issue #152, opened 2026-05-05
Test interaction-aware pCR prediction with regularized XGBoost after near-duplicate feature pruning


RebeccaRuochunWu:
## Why this issue exists

  We currently have ~1k engineered tabular features across clinical, tumor-size, morphometry, graph, kinematic, and second-order feature blocks. Many of these features are
  intentionally related: raw counts, shell counts, densities, normalized ratios, log transforms, and shell contrasts often describe the same underlying vessel/tumor
  measurement in several ways.

  The current elastic-net feature-selection pipeline is useful as a leakage-aware marginal/linear baseline. However, it can miss interactions such as:

  - feature A is high and feature B is low
  - near-tumor enhancement is high only when vessel burden is low
  - a graph feature matters only for large tumors

  XGBoost is a good next model to test because shallow trees can learn threshold-based feature interactions. But before using XGBoost, we should remove obvious near-
  duplicate features so the model is not spending capacity on many versions of the same measurement.

  ## Main question

  Does a regularized XGBoost model, trained on a near-duplicate-pruned feature set, improve held-out pCR prediction compared with the current elastic-net feature-selection
  pipeline?

  ## Important principle

  Do **not** use Mann-Whitney/BH screening before XGBoost.

  That screen is univariate and label-based. It can remove features that only matter jointly with other features. For this issue, XGBoost should get features after
  **unsupervised cleanup only**.

  ## Feature blocks to consider

  Current feature blocks are:

  - `clinical`: non-imaging metadata
  - `tumor_size`: tumor voxel counts and peritumoral shell-size summaries
  - `morph`: whole-network vessel morphometry summaries
  - `graph`: tumor-centered structural vessel features
  - `kinematic`: tumor-centered dynamic vessel features
  - second-order features derived from the above blocks

  The pruning should be aware of these blocks. Do not just compute one global correlation matrix and drop everything above a threshold without considering what the
  features mean.

  ## Near-duplicate pruning plan

  ### Step 1: Basic unsupervised cleanup

  Inside each training fold, remove:

  - all-NaN features
  - low-coverage features, e.g. non-missing in `<50%` of training cases
  - near-constant features
  - exact duplicate columns
  - columns with only one non-missing value or one unique value

  This must be done using the training fold only.

  ### Step 2: Identify near-duplicate correlated groups

  Use a high threshold, not `0.8`.

  Suggested starting rule:

  - compute correlations using training-fold data only
[...]

PR #155, opened 2026-05-05
 Implements XGBoost interaction-aware pCR prediction pipeline as a nonlinear comparator to the elastic-net baseline 

RebeccaRuochunWu:
## Summary
   Implements XGBoost interaction-aware pCR prediction pipeline as a nonlinear comparator to the elastic-net baseline (related to #152).

   ## What's included
   - **`scripts/xgboost_interaction_pipeline.py`**: Full pipeline with block-aware near-duplicate pruning, XGBoost (GPU) +
  elastic-net arms, 15-fold CV, 8 visualizations
   - **`slurm/submit_xgboost_interaction.slurm`**: GPU-enabled Slurm submission script

   ## Results
   | Metric | XGBoost | Elastic-net |
   |--------|---------|-------------|
   | AUC | 0.581 ± 0.060 | 0.598 ± 0.060 |
   | AP  | 0.388 ± 0.052 | 0.416 ± 0.049 |

   Neither model shows strong signal. Elastic-net slightly edges XGBoost, suggesting current features may lack strong
  interaction effects for pCR prediction. Top XGBoost feature: `per_shell_topology_shell_10_20mm_shell_region_volume_mm3`.

image
image
image
image
image


annawoodard (approved):
Thanks for putting this together — nice PR. The pipeline reads cleanly, framing it as interaction-aware XGBoost vs. the marginal/linear elastic-net baseline is the right comparison, and having the outputs committed makes it easy to poke at the results. Your PR writeup was very nice and all the plots are useful/readable/help tell the story. Also glad you kept the interpretation conservative given XGBoost doesn't actually beat the baseline here.

Approving pending two small path fixes:

1.  `slurm/submit_xgboost_interaction.slurm.py`: the default `INPUT_CSV` has `vanguard/...` in it, but we submit from the repo root and the script cds there, so the default ends up resolving to `<repo>/vanguard/experiments/`.... Should just be experiments/second_order_features/features_engineered_labeled.csv.
2. `scripts/xgboost_interaction_pipeline.py`: the docstring shows `python scripts/xgboost_interaction_pipeline.py` ..., but that fails on its own because the script imports features and needs the repo root on `PYTHONPATH`. The Slurm script already sets it, so either change the usage line to `PYTHONPATH=. python` ... or drop a small path shim at the top of the script. Here's an example

```
 import sys
  from pathlib import Path

  REPO_ROOT = Path(__file__).resolve().parents[1]
  if str(REPO_ROOT) not in sys.path:
      sys.path.insert(0, str(REPO_ROOT))
```
  Put that before:
```
  from features import feature_block_for_column
```

Thanks for your great work on this!

PR #148, opened 2026-04-29
Deep Sets: add inclusion-rule comparison and set default

danielluna90:
Changes
- Add configurable inclusion-rule comparison in builder and emit inclusion_rule_summary.csv.
- Keep local_radius_with_fallback as default, with documented rationale and fixture stats.
- Add tests and config/docs/slurm updates for reproducible rule comparison.

Follow-up:
- Fixture stats validate behavior and decision logic, but full-cohort runtime stats should be generated in cluster Slurm runs and attached to the PR for production-scale confirmation.

PR #156, opened 2026-05-11
Deep Sets: compare fixed-radius and peritumoral inclusion variants (#121)

### LINKED ISSUE #121: Compare tumor-local inclusion rules for Deep Sets point sets
**Why this issue exists**

Which points you include can matter as much as which features you attach to them. A set that is too wide may include lots of irrelevant vessel points. A set that is too narrow may miss useful peritumoral structure.

**Main goal**

Compare a small set of inclusion rules and choose one default rule for later Deep Sets modeling.

**What the current rule does**

The current rule is tumor-relative:

- compute `tumor_equiv_radius_mm`
- set `local_radius_mm = max(floor, scale * tumor_equiv_radius_mm)`
- cap if needed
- keep points within that radius from the tumor boundary

**What to compare**

Keep this small. A good starting set is:

- current tumor-relative radius rule
- fixed-radius rule, for example `30 mm`
- fixed-radius rule, for example `50 mm`
- tighter peritumoral rule, such as inside tumor plus first few shells

Optional extra:

- nearest-`N` rule as a controlled fallback-style baseline

**What to measure**

At minimum:

- number of cases written
- number of skipped cases
- fraction of fallback cases
- median and range of `num_points`

Good optional downstream check:

- run one small baseline Deep Sets training job for each inclusion rule

For anything beyond a tiny smoke check, use the existing Deep Sets Slurm pipeline.

**friendly advice**

Do not judge an inclusion rule only by "more points seems better."

Also ask:

- does this rule keep a reasonable number of points?
- does it avoid frequent fallback behavior?
- does it focus on the tumor-local region we actually care about?

**Definition of done**

- a few inclusion rules have been compared
- there is a compact summary table of set-build statistics
- one default inclusion rule is chosen for the Deep Sets Modeling and Evaluation Subproject
- the reasoning for that choice is written down

**Helpful file pointers**

- `build_deepsets_dataset.py`
- `slurm/submit_deepsets_pipeline.sh`
- `slurm/deepsets_job.slurm`

danielluna90:
**Description**
Closes the remaining scope from [#121](https://github.com/dsi-clinic/vanguard/issues/121) after #148: compare fixed-radius cutoffs and a tighter peritumoral-shell rule alongside the existing tumor-relative inclusion presets.

**What changed**

- build_deepsets_dataset.py now supports three additional comparison rules:
  - fixed_radius_30mm_with_fallback
  - fixed_radius_50mm_with_fallback
  - peritumoral_shells_with_fallback (inside tumor plus the first outside shells up to 5 mm)
 
- Each compared rule still reports the same build stats in `inclusion_rule_summary.csv`: cases written, cases skipped, fallback fraction, and point-count median/range.
- Default Deep Sets configs and the issue #121 fixture config include the new variants in deepsets_compare_inclusion_rules.
- `analysis/deepsets_issue121_notebook.ipynb` documents the expanded comparison set and prints the new summary rows.
- Unit tests cover cutoff helpers and per-case comparison metrics.

**Default rule**
`local_radius_with_fallback` remains the default. The new variants are comparison-only presets for build stats and notebook review.


jpivarski (approved):
Looks good! It satisfies the remainder of #121, and I checked it over and don't see any issues. Feel free to squash-and-merge.

Issue #118, opened 2026-04-02
Summarize the best tabular model against the baseline arms


annawoodard:
**Why this issue exists**

Even after choosing the best tabular model setup, we still need to explain what that model is buying us over simpler baselines.

The Tabular Modeling and Evaluation Subproject should end with a clean comparison against the main baseline arms, not just an isolated "best model" number.

**Main goal**

Use the chosen tabular-modeling setup to summarize results against the baseline arms in a way that is easy for the team to read and reuse later.

**What to compare**

At minimum:

- `clinical + tumor_size`
- `clinical + tumor_size + morph`
- `clinical + tumor_size + graph`
- `clinical + tumor_size + kinematic`
- `clinical + tumor_size + morph + graph + kinematic`
- strongest available radiomics baseline for the same cohort, or the closest matched radiomics experiment if the exact setup differs

These are already close to the independent-signal setup in:

- `configs/independent_signal.yaml`

The radiomics baseline lives in the separate radiomics workflow under:

- `radiomics/`

**What to do**

1. Freeze the chosen model family and tuning approach from #116 and #117. 
2. Reuse or extend the independent-signal config.
3. Rerun the vessel-feature baseline-arm comparison with the chosen tabular-modeling setup.
4. Pull in the strongest available radiomics baseline on the same cohort if possible.
5. If the radiomics setup is not exactly matched, document the mismatch clearly:
   - cohort
   - splits
   - labels
   - any major preprocessing differences
6. Write a short results note with:
   - one table of vessel-feature arm results
   - one row or separate table for the radiomics baseline
   - delta vs `clinical + tumor_size`
   - one paragraph explaining which vessel arms helped most and how they compare with radiomics

**What a good final summary looks like**

A good summary should answer:

- which single vessel block helped the most?
- does combining all vessel blocks help more than any one block alone?
- is the gain over `clinical + tumor_size` large enough to care about?
- how does the best vessel-feature tabular model compare with the radiomics baseline?
- does the conclusion still match what we saw in the Tabular Feature Engineering Subproject?

**Definition of done**

- the baseline-arm comparison has been rerun with the chosen tabular-modeling setup
- the summary includes a radiomics baseline comparison
- there is a compact result table and short write-up
- the write-up is understandable without reading the code

**Helpful file pointers**

- `configs/independent_signal.yaml`
- `run_ablation_matrix.py`
- `results/independent_signal_q3_summary.csv`
- `slurm/submit_feature_ablation.slurm`
[...]

PR #157, opened 2026-05-11
Audit tabular feature coverage across all 4 datasets; add multi-dataset config (Issue #150)

### LINKED ISSUE #150: Verify full tabular feature coverage across datasets; rerun Slurm jobs as needed
## Summary

Audit how much of our pCR-labeled cohort actually has complete tabular features
(clinical, tumor_size, morph, graph, kinematic), produce a per-dataset
coverage report, and rerun any upstream Slurm jobs needed to close the gaps so
future tabular runs (e.g. #118 follow-ups) are not stuck on ISPY2 only.

## Motivation

The #118 baseline-arm comparison was evaluated on **n = 808** ISPY2 unilateral
cases. We have pCR labels for **1491 cases** across four datasets, so roughly
**45% of the labeled cohort is not currently usable** by the tabular pipeline:

| Dataset | pCR-labeled cases | Used in #118 |
|---|---|---|
| ISPY1 | 167 | 0 |
| ISPY2 | 980 | 808 (unilateral, full features) |
| DUKE | 280 | 0 |
| NACT | 64 | 0 |
| **Total** | **1491** | **808** |

We do not currently know whether the gap is missing labels, missing
segmentations / centerlines, missing tumor-graph JSONs, missing kinematic
features, or just missing config / cohort code. This issue is the audit + the
rerun work to fix whichever of those is the actual bottleneck.

## Scope

Datasets: **ISPY1, ISPY2 (unilateral + bilateral), DUKE, NACT**.

Tabular feature blocks audited per case:
- `clinical`
- `tumor_size`
- `morph`
- `graph`
- `kinematic`

Plus the upstream artifacts each block depends on:
- vessel segmentation
- centerline `.npy`
- tumor mask
- tumor-graph JSON

### 1. Build a coverage table

One CSV per dataset with one row per case_id and a column per artifact /
feature block, recording present / missing. Aggregate into a summary like:

| Dataset | Has label | Has segmentation | Has centerline | Has tumor mask | Has tumor-graph | Has all 5 feature blocks |
|---|---|---|---|---|---|---|
| ISPY1 | 167 | ? | ? | ? | ? | ? |
| ISPY2 | 980 | ? | ? | ? | ? | ? |
| DUKE | 280 | ? | ? | ? | ? | ? |
| NACT | 64 | ? | ? | ? | ? | ? |

Saved to `results/feature_coverage_audit.csv` so the numbers are tracked.

### 2. Identify which Slurm jobs need to be rerun

Map each missing artifact to the existing pipeline step:
- vessel segmentation → `segmentation/slurm/submit_batch_segmentation_array.slurm`
- centerline / tumor-graph → `graph_extraction/slurm/submit_tc4d_array.slurm`
- per-block tabular features → call sites in `features/`

### 3. Run gap-filling jobs (only what's missing)

Submit Slurm arrays scoped to the missing case_ids, not full reruns. Keep logs
under `logs/feature_coverage_<dataset>-<jobid>.out` so reruns are auditable.

### 4. Extend cohort / configs to consume the new datasets

- Add `configs/ispy1.yaml`, `configs/duke.yaml`, `configs/nact.yaml` mirroring
  `configs/ispy2.yaml`, with the right `data_paths` and label paths.
- Confirm `tabular_cohort.py` handles the union case — `dataset_include` already
  accepts a list, but it has only ever been exercised with a single dataset.

### 5. Re-audit and update the coverage table

After rerunning, regenerate `results/feature_coverage_audit.csv` and document
what changed.

## Deliverables

- `results/feature_coverage_audit.csv` (before + after the reruns)
- New per-dataset configs under `configs/`
- A short note in `docs/` summarizing what was missing, what we reran, what
  remains genuinely unrecoverable (e.g. studies that never had usable imaging),
  and the new "max usable n" per dataset for downstream tabular work
- Slurm logs in `logs/`

## Definition of done

- We can answer, per dataset, exactly how many cases have all 5 tabular feature
  blocks computed.
- Anything that *can* be filled in has been filled in via Slurm reruns.
- A multi-dataset config (or the ability to set `dataset_include: [ISPY1,
  ISPY2, DUKE, NACT]`) actually loads end-to-end without the cohort builder
  silently dropping rows.
- The coverage table is committed and linked from the relevant docs.

## Open questions for @annawoodard

1. Should we keep `bilateral_filter: False` for the ISPY2 audit, or also try to
   recover the bilateral cases now that we are widening the cohort anyway?
2. For DUKE specifically — is exclusion a data-quality decision we want to keep
   (per past discussion), or is the original exclusion just because
   segmentation / centerline runs were never done for it?
3. Is there a dataset priority order if cluster time is limited? My default
   would be DUKE > ISPY1 > NACT (largest first).
4. Where should the new per-dataset configs live? Same `configs/` directory, or
   a `configs/cohorts/` subfolder?

## Refs
- Refs #118 (current ISPY2-only constraint comes from there)

danielgong1:
…et config

Implements Issue #150. The audit confirms that the upstream tabular pipeline has already been run for every study directory across DUKE, ISPY1, ISPY2, and NACT (1506 / 1506 have centerline .npy, morphometry JSON, tumor-graph JSON with status=ok, and tumor mask), so no Slurm reruns are required. The "only 808 used in #118" gap was entirely a config artifact (dataset_include
+ bilateral_filter), not a data gap. With those filters relaxed via the new configs/all_datasets.yaml, tabular_cohort.py loads the full 1491-case labeled cohort (DUKE 280 + ISPY1 167 + ISPY2 980 + NACT 64) end-to-end with no code changes.

Adds:
- scripts/audit_feature_coverage.py: reproducible read-only audit walking the centerline tree and the tumor-mask / pcr-label / radiomics-label indices, emitting per-case + per-dataset CSVs.
- results/feature_coverage_audit.csv: per-case audit table (1506 rows).
- results/feature_coverage_summary.csv: per-dataset summary.
- configs/all_datasets.yaml: multi-dataset cohort config (dataset_include for all four datasets, bilateral_filter null).
- docs/issue150_feature_coverage_audit.md: background, key numbers, smoke test trace, and reproduction instructions.

Closes #150
Refs #118 (the ISPY2-only constraint this audit removes)


annawoodard (approved):
Nice work on this. The audit is well-scoped and reproducible, and the summary matched when I reran it. Good call explaining *why* we hit 808 cases (config filters, not missing upstream centerline artifacts).

One cleanup before merge: make the clinical metadata source explicit and actually audit it.

`configs/all_datasets.yaml` sets both `patient_info_dir` and `clinical_excel`, but `clinical_features.get_clinical_features()` uses `patient_info_dir` first and only falls back to `clinical_excel` if the JSON directory is missing. So the pipeline is actually loading patient-info JSON files, not Excel files. That's fine, no need to switch, but the audit and doc should match that reality.

Concrete changes:

1. In `scripts/audit_feature_coverage.py`, add a `--patient-info-dir` argument, defaulting to `/net/projects2/vanguard/MAMA-MIA-syn60868042/patient_info_files`.
2. For each audited `case_id`, check whether `<patient_info_dir>/<case_id>.json` exists (adjust the filename pattern if it's different in the repo).
3. Add a per-case column `has_patient_info_json` to `results/feature_coverage_audit.csv`.
4. Add a summary column `n_patient_info_json` to `results/feature_coverage_summary.csv`.
5. Include `has_patient_info_json` in the `has_all_artifacts` conjunction, since `use_clinical: true` is set in the config.
6. In `docs/issue150_feature_coverage_audit.md`, update the clinical-source wording to say patient-info JSONs are the actual source, with Excel only as a fallback.
7. Rerun the audit and update both result CSVs.

Overall this is solid; the cleanup just makes the audit match the path the model will actually use. Ping me if anything's unclear. I'm approving this to keep things moving.

PR #159, opened 2026-05-13
Ablation study - adding different groups of first order & derived features to baseline clinical features and run elastic net

RebeccaRuochunWu:
Summary

Ablation study evaluating the incremental contribution of MRI-derived feature blocks to pCR prediction, using clinical features as baseline.

What's included

 - Configs: clinical_graph_ablation.yaml (6-arm ablation matrix), ispy2_with_clinical.yaml (feature engineering with clinical 
predictors)
 - Scripts: Updated feature_selection.py and xgboost_interaction_pipeline.py
 - Slurm: Submission scripts for ablation and feature selection jobs

Ablation Results (5-fold CV, n=808)

| Arm | # Features | AUC (mean +/- std) |
 |-----|------------|---------------------|
 | Clinical only | 4 | 0.480 +/- 0.017 |
 | + Tumor size | 17 | 0.590 +/- 0.064 |
 | + Morph | 67 | **0.606 +/- 0.045** |
 | + Graph (no kin) | 121 | 0.594 +/- 0.033 |
 | + Kinematic (no graph) | 894 | 0.583 +/- 0.046 |
 | + All vessel | 948 | 0.580 +/- 0.049 |

image

Key findings

 - Tumor size provides the largest single uplift (+0.11 AUC over clinical alone)
 - Morphometry adds modest additional value (best arm at
  0.606)
 - High-dimensional graph/kinematic features don't improve over morph alone, likely due to overfitting with top-k=64 feature 
selection

 Experiment outputs not committed — reproducible via Slurm scripts.


annawoodard (approved):
Nice work on this!!! The XGBoost vs elastic-net comparison is set up cleanly, feature selection stays inside the CV folds (good!), and the fold-level summaries make everything auditable. Also appreciate that you reported the result honestly: elastic-net coming out slightly ahead is useful evidence either way.

Two things to fix before merge:

**1. Near-duplicate pruning is chaining features.**

`scripts/xgboost_interaction_pipeline.py` currently builds connected components from pairwise correlations and keeps one feature per component. The issue: A↔B and B↔C can both be above threshold while A↔C isn't — so you end up dropping features that aren't actually near-duplicates of what replaced them. You can see this in `feature_pruning_report.csv`, where some dropped features have low correlation to the kept feature.

Switch to a greedy rule instead: sort candidates by the representative score, keep the best remaining feature, and only drop others whose absolute correlation to *that kept feature* is above the threshold. Then every dropped feature is genuinely a near-duplicate of its replacement.

After fixing, rerun and update:
- `experiments/xgboost_interaction/feature_pruning_report.csv`
- `experiments/xgboost_interaction/fold_summaries.csv`
- `experiments/xgboost_interaction/comparison_table.csv`
- related plots

**2. Default input path in `slurm/submit_xgboost_interaction.slurm` is wrong.**

Current:
```bash
INPUT_CSV="${INPUT_CSV:-vanguard/experiments/second_order_features/features_engineered_labeled.csv}"
```
Submitted from the repo root, this resolves to `vanguard/vanguard/...`. Change to:
```bash
INPUT_CSV="${INPUT_CSV:-experiments/second_order_features/features_engineered_labeled.csv}"
```
Or make it absolute.

**While you're regenerating plots:** please clean up extraneous color. Bar plots where each bar is a different color for no reason are a pet peeve of mine. Color should encode information, not just decorate. If a bar's color doesn't mean anything (a category, a grouping, highlighting one bar against the rest), use a single neutral color for all of them. Same goes for line plots, scatter plots, etc. Tufte's *The Visual Display of Quantitative Information* is the canonical reference here if you want to go deeper. The short version is that every visual element should be doing work.

**Optional, not a blocker:** when `--site-group-cv` is used, the outer CV is site-grouped but inner hyperparameter tuning still uses standard stratified CV. That's fine for now — just add a note in the script help text or a comment so we're not overstating what grouped-CV mode does.

Overall this is a good experiment and a useful result. The main thing is getting the pruning right so the comparison reflects actual near-duplicate removal. Ping me if anything's unclear!

Approving to keep things moving, I trust you to put these changes in and merge without me re-reviewing it.

annawoodard:
I refreshed the XGBoost interaction artifacts after the pruning/launch fixes.

  What changed:

  - Replaced connected-component near-duplicate pruning with greedy representative pruning, so every dropped feature is directly correlated with its kept representative at
    >= 0.97.
  - Fixed the Slurm default input path and added a missing-input guard.
  - Reran the artifact generation with the clinical-enhanced input and --encode-clinical.
  - Cleaned stale PDP outputs so the committed PDP plots match the current rerun.

  Why the results changed: the previous XGBoost artifacts did not actually include encoded categorical clinical variables. Raw fields like tumor_subtype /
  menopausal_status are metadata unless --encode-clinical is used. The refreshed run includes 8 clinical candidate features.

  | Arm | # Features | AUC mean +/- std | AP mean +/- std |
  |---|---:|---:|---:|
  | Previous XGBoost, no encoded clinical | 1,179 candidates; 445 post-pruning avg | 0.581 +/- 0.060 | 0.388 +/- 0.052 |
  | Refreshed XGBoost, encoded clinical | 1,215 candidates; 536 post-pruning avg | 0.660 +/- 0.032 | 0.447 +/- 0.039 |
  | Previous elastic-net, no encoded clinical | 1,179 candidates; 8 selected avg | 0.598 +/- 0.060 | 0.416 +/- 0.049 |
  | Refreshed elastic-net, encoded clinical | 1,215 candidates; 6 selected avg | 0.684 +/- 0.034 | 0.468 +/- 0.044 |

  Validation: the refreshed pruning report has 7,426 dropped-feature rows, minimum correlation_to_kept = 0.97, and zero threshold violations.

  Takeaway: the refreshed run is much stronger than the previous artifacts, but the improvement appears mostly driven by clinical covariates. Elastic-net still slightly
  outperforms XGBoost, and the top XGBoost signal is clinical_luminal_a, so I would frame this as a cleaner interaction-aware comparison rather than evidence that vessel
  interactions are driving performance.

PR #160, opened 2026-05-13
Ablation study - Add XGBoost ablation for clinical graph feature evaluation

RebeccaRuochunWu:
Summary

  Extends the clinical graph ablation study by adding an XGBoost (nonlinear) model alongside the existing Elastic Net (linear)
  baseline. The goal is to test whether tree-based models can better leverage graph and kinematic vessel features through nonlinear
  interactions.

  What's included

   - configs/clinical_graph_ablation_xgb.yaml — XGBoost ablation config (n_estimators=300, max_depth=4, learning_rate=0.05)
   - Same 6-arm ablation design as the Elastic Net study
   - Reuses the existing slurm/submit_clinical_graph_ablation.slurm with config override

  Results: Elastic Net vs XGBoost (5-fold CV, n=808)

 | Arm | Features | LR AUC | XGB AUC |
 | --- | --- | --- | --- |
 | Clinical only | 4 | 0.480 | 0.465 |
 | + Tumor size | 17 | 0.590 | 0.520 |
 | + Morph | 67 | **0.606** | 0.585 |
 | + Graph (no kin) | 121 | 0.594 | 0.587 |
 | + Kinematic (no graph) | 894 | 0.583 | 0.579 |
 | + All vessel | 948 | 0.580 | 0.583 |

image

  Key findings

   1. Elastic Net outperforms XGBoost across most arms, particularly for lower-dimensional feature sets (clinical only, + tumor size).
   2. Both models show the same pattern: morphometry is the performance peak; adding graph/kinematic features does not help.
   3. XGBoost narrows the gap at higher dimensions — the + Graph and + All Vessel arms are nearly equivalent across models, suggesting
   XGBoost handles collinearity better but still can't extract useful signal from the high-dimensional features.
   4. Likely explanations for XGBoost underperformance:
    - The top-k=64 univariate feature selection pipeline was designed for linear models and may filter out features valuable for 
  tree-based interactions.
    - With n=808 samples and shallow trees, XGBoost may lack sufficient data to learn meaningful nonlinear interactions.

  Possible next steps

   - Run XGBoost without top-k filtering (let the tree handle feature selection natively)
   - Increase feature_select_k for XGBoost arms
   - Try nested CV hyperparameter tuning for XGBoost


annawoodard (approved):
Great work!

Some changes:

**1. Make the committed artifacts match the PR claim.**

The PR description reports the 6-arm XGBoost ablation, but the files in `experiments/xgboost_interaction/*` come from `scripts/xgboost_interaction_pipeline.py`, which is a different analysis. Pick one:
- Commit the actual `run_ablation_matrix.py` outputs for the 6-arm run: `ablation_summary.csv` and `ablation_fold_auc.csv` if you have it.
- Or remove the `experiments/xgboost_interaction/*` artifacts and keep this PR focused on the ablation configs.

**2. Add the command used to generate the reported table.**

Drop the exact `run_ablation_matrix.py --config ... --outdir ...` invocation somewhere visible — PR body or a short note next to the output files is fine. Just needs to be runnable as-is.

**3. Fix the Slurm default input path.**

In `slurm/submit_xgboost_interaction.slurm:14`, the default is `vanguard/experiments/...`, and after `cd "${REPO_ROOT}"` that resolves to `vanguard/vanguard/...`. Change to:
```bash
INPUT_CSV="${INPUT_CSV:-experiments/second_order_features/features_engineered_labeled.csv}"
```
Or make it absolute. Also add a quick check that the file exists before the job kicks off — fails fast and saves a Slurm slot.

**4. Make the plot script safer.**

In `scripts/plot_model_auc_comparison.py`, missing model/arm rows currently get plotted as AUC 0, which makes it look like there's a real result when there isn't. Have it fail with a clear error message instead. Also pull out the hard-coded "largely linear" conclusion text — interpretation belongs in the PR or a result note, not baked into the plot script.

**Optional if there's time:**

The elastic-net comparator inside `scripts/xgboost_interaction_pipeline.py` doesn't fully reuse `feature_selection.py`'s preprocessing, so I'd avoid calling it "the baseline" unless those are aligned. The easiest fix is to soften the wording to "exploratory elastic-net comparator" rather than implying it replicates the full baseline pipeline.

I'm approving now to keep things moving. I trust you to fix the items above and merge when ready. Ping me if anything's unclear!

Issue #151, opened 2026-05-01
Evaluate top features vs tumor-size baseline; build reproducible, modular analysis framework


danielgong1:
## Summary

We have shown (in #118) that adding the full vessel feature stack lifts AUC over
the `clinical + tumor_size` baseline. We have **not** shown which individual
features inside that stack are doing the work, or whether a small handful of top
features is enough to recover most of the gain. This issue is to build a
reproducible, modular framework that:

1. ranks individual tabular features by predictive signal,
2. evaluates top-K features against the `clinical + tumor_size` baseline using
   the same evaluation regime as #118, and
3. produces an artifact / report that future analyses can reuse without
   rewriting plumbing.

## Motivation

- The current selectors (`NumericTopKAUCSelector`, `NumericKinematicAddonSelector`
  in `tabular_models.py`) pick features per-fold, but their picks are not
  surfaced anywhere downstream — we cannot answer "which features were chosen
  most often across folds" today without re-instrumenting.
- Block-level ablation (#118) tells us morph + graph + kinematic together help,
  but does not separate "this single kinematic feature carries most of the
  signal" from "the lift is spread across many features."
- A modular framework also lets us reuse the same rank → evaluate → report
  pipeline for future feature engineering work (e.g., when second-order
  features land, or when #150's expanded cohort is ready).

## Scope

- **Datasets:** ISPY2 unilateral (n = 808) for the initial run, to stay
  comparable with #118. The framework should be dataset-agnostic so it can be
  rerun on the multi-dataset cohort once #150 lands.
- **Models:** logistic regression (nested-tuned) and XGBoost — the carry-forward
  pair from #116 / #117.
- **Baseline:** `clinical + tumor_size` (LR), per #118.
- **Feature universe:** all numeric tabular features available across the five
  blocks (`clinical`, `tumor_size`, `morph`, `graph`, `kinematic`).

### 1. Ranking layer — per-feature predictive signal

Two ranking methods, both applied within nested CV (so the ranking is computed
on outer-train rows only, not the held-out fold):

- **Univariate AUC** — per-feature AUC against pCR on the outer-train rows.
- **Model-based importance** — XGBoost gain or LR absolute standardized
  coefficient, fit on outer-train rows.

Both rankings are saved per fold for stability analysis (Jaccard overlap of
top-K across folds).

### 2. Evaluation layer — top-K vs. baseline

- For each ranking method × K ∈ {5, 10, 20, 50, "all"}, run a #118-style
[...]

PR #161, opened 2026-05-19
Top-feature ranking workflow for baseline vs top-K arms (Issue #151)

danielgong1:
## Summary
- Adds `analysis/feature_ranking.py` (univariate AUC, XGB gain, LR |coef|) and `analysis/top_k_arms.py` to assemble baseline + explicit top-K arms from a global ranking table.
- Extends `tabular_cohort.select_features` and `run_ablation_matrix` / modeling helpers with optional `explicit_model_columns` so arms can subset columns without new runners.
- Adds `run_top_features_eval.py`, `configs/top_features_eval.yaml`, `slurm/submit_issue151_top_features.slurm`, and `docs/issue151_top_features_eval.md`.
- Registers `top_features_eval` in default config; adds `pytest` to `environment.yml` with tests under `tests/`.
## Test plan
- [ ] `pytest tests/test_feature_ranking.py tests/test_top_k_arms.py`
- [ ] Optional cluster smoke: `sbatch slurm/submit_issue151_top_features.slurm` from repo root with usable data paths.
Refs #151.
Important


annawoodard (approved):
Thanks, this is really nice work. The structure is clean: separate ranking utilities, separate top-K arm construction, a dedicated runner, a worked config, docs, Slurm entrypoint, and focused tests. I also appreciate that the docs explicitly say the global ranking is exploratory/optimistic and shouldn't be treated as leakage-safe inference. That's exactly the right framing.

I'm approving now to keep things moving. I trust you to fix the items below and merge when ready.

`run_top_features_eval.py` uses `--outdir` before ensuring it exists. It calls `_prepare_full_dataset(config, placeholder_arms, outdir)`, and `_prepare_full_dataset()` writes `features_full_raw.csv` directly into `outdir`. Since the example command uses a fresh directory like `experiments/issue151_run`, a first-time run will fail unless the user creates it manually.

Add this before calling `_prepare_full_dataset()`:
```python
outdir = args.outdir
outdir.mkdir(parents=True, exist_ok=True)
```


Ping me if anything's unclear!

PR #162, opened 2026-05-20
Add quantitative justification for block-aware feature selection

RebeccaRuochunWu:
## Summary
 
 Adds an analysis script and documentation justifying why feature selection
 should be block-aware rather than global.
 
 ## Key findings (from running on the vessel_all feature table)
 
 - **946 numeric features** across 5 blocks, with kinematic alone at 87% (827 features)
 - Intra-block mean |ρ| = 0.420 vs inter-block mean |ρ| = 0.152 — blocks are internally redundant but mutually distinct
 - Global top-64 **eliminates clinical entirely** and gives kinematic 67% of slots
 - Block-aware selection preserves all blocks
 
image

image

 ## Files
 
 - `scripts/analyze_block_selection_bias.py` — produces three diagnostics from a labeled feature CSV
 - `docs/block_feature_selection_justification.md` — explains the problem and methodology


annawoodard (approved):
Thanks Rebecca, great work! This is a useful diagnostic, and the structure is clear: block sizes, within/between-block correlation, and global-vs-block-aware survival are exactly the right checks for this question.

I'm approving now to keep things moving — I trust you to make the changes below and merge when ready.

**Before merge:**

1. The doc is titled as a quantitative justification, but right now it mostly describes the *expected* result. Run the script on the intended feature table and paste the actual key outputs into the doc: the block size table, the intra-vs-inter-block correlation summary, and the global vs block-aware survival table. That'll make the argument concrete instead of hypothetical.

2. In `scripts/analyze_block_selection_bias.py`, swap the manual AUC calculation for `sklearn.metrics.roc_auc_score`. The current pairwise comparison handles ties incorrectly, which can shift feature rankings. Use `max(auc, 1 - auc)` if you want direction-invariant ranking, and skip features whose finite labels don't contain both classes.

**Minor:** while you're updating the doc, double-check that any config names you reference actually exist in the repo.

Ping me if anything's unclear!

PR #163, opened 2026-06-26
Restructure repo

spencervenancio:
## What?
Reorganize repository into per-family subpackages and consolidate environment files

## Why?
The repo root was too flat — model files, scripts, and config were all at the top level with no clear ownership. Also discovered a Randi-specific install issue with `pyradiomics` that needed documenting.

## How?
- Moved DeepSets files to `vanguard/deepsets/` and tabular modeling files to `vanguard/tabular/`; updated all imports accordingly
- Merged `requirements.txt` into `environment.yaml` (single source of truth for dependencies)
- Added a `README.md` note: on Randi, `pyradiomics` has no prebuilt wheel for Python 3.10 and will fail if `/tmp` is noexec — fix is to export `TMPDIR` to a writable scratch path before installing

## Testing?
- Verified imports work after the move
- Confirmed the environment install instructions in the README complete successfully


annawoodard (changes requested):
Thanks for taking this on. Great first pass!! The direction is good: moving Deep Sets, tabular modeling, and modeling runners into clearer subpackages should make ownership easier for new contributors.

Before merging, this needs a cleanup pass focused on entrypoint reliability. For a repo restructure, the main review question is not just "do imports work?", but "can a user still copy-paste the documented commands and can Slurm still run the pipeline from a clean shell?"

Changes needed:
1. Some moved Deep Sets scripts are now called by file path from Slurm, e.g. `python deepsets/merge_deepsets_manifest.py`, while the files import from the `deepsets` package. In a clean  shell without `PYTHONPATH` set, that can fail because the repo root is not on `sys.path`. I reproduced this with:
     `env -u PYTHONPATH python deepsets/merge_deepsets_manifest.py --help`
which gives `ModuleNotFoundError: No module named 'deepsets'`. Please either switch these calls to module form, e.g. `python -m deepsets.merge_deepsets_manifest`, `python -m deepsets.build_deepsets_dataset`, `python -m deepsets.train_deepsets`, or add the repo root to `sys.path` consistently in the moved CLI files.
  2. Several docs/scripts still reference deleted top-level entrypoints such as `train_tabular.py`, `train_deepsets.py`, `build_deepsets_dataset.py`, `merge_deepsets_manifest.py`, and
  `run_ablation_matrix.py`. Examples include the main README, `evaluation/README.md`, `slurm/README.md`, `scripts/benchmark_deepsets_feature_sets.sh`, and `scripts/
  issue121_fixture_smoke.sh`. 
3. The new Randi README block has a malformed code fence: `source ~/.bashrc```` is on the same line as the closing fence. That will make the installation instructions render incorrectly.
4. The tests (which is basically linting IIRC) are not passing. It looks like somewhere along the way we lost the documentation for running those locally before pushing. I'll dig them up and send them to you on Slack, let's add them back into the readme.


Finally, this is just a preference, but I favor removing redundancy in filepath names. E.g. `deepsets/deepsets_model.py` -> `deepsets/model.py`. This reduces cognitive overhead when screening paths.

spencervenancio:
Thanks for the thorough review. This pass focuses on entrypoint reliability from a clean shell. Responses to each point:

1. File-path Slurm calls failing without PYTHONPATH. 
Fixed at the root by making the package editable-installable (pip install -e is now in environment.yml) AND ensuring the install actually exposes everything the moved scripts import. The subpackages alone weren't enough: the scripts also import top-level modules (clinical features, load_ cohort, config, clinic metadata) and the features/ and analysis/ packages. pyproject.toml now declares these via py-modules and the package include list, so file-path invocation works in a clean shell. Verified:
`env -u PYTHONPATH python deepsets/build_dataset.py --help` and `env -u PYTHONPATH python -m pytest tests/ -q` (76 passed)

2. Stale top-level entrypoint references 
Audited every python <path>•py reference in docs/scripts/slurm/configs; all resolve. run_ablation matrix-py references now point at modeling/run_ablation matrix.py; no references to the removed deepsets/tabular top-level names remain.
3. Malformed Randi code fence fixed; the install block now renders correctly.
4. Checks not passing + lost local-run docs _ pre-commit run --all-files passes.
5. Redundant filename prefixes applied (deepsets/deepsets_model.py -> deepsets/model.py, etc.).

annawoodard (approved):
Fabulous!!!! Thank you Spencer!!! This is a huge improvement

👏 👏 👏 👏 👏 👏 👏 👏

PR #167, opened 2026-06-30
Add faster segmentation experiment and pipeline inventory tracking

SBose-12:
## Summary
Adds a standalone faster segmentation experiment alongside the existing segmentation pipeline, plus a new inventory script for tracking pipeline completion status.

## What's included
- `faster-segmentation-test/` — speedup-optimized segmentation script (`batch_segmentation_fast.py`, `predict_fast.py`) with output comparison validation (`validate_outputs.py`) and equivalence tests (`test_batching_equiv.py`, `test_preprocess_parallel.py`)
- `scripts/inventory.py` — tracks completion status across segmentation, centerline, and feature extraction stages
- Array job submission scripts (`submit_fast_array.sh`, `submit_fast_array.slurm`) for batch processing on the cluster

## Scope
This is purely additive — nothing in the existing segmentation pipeline is modified or replaced. The fast version runs as a separate, parallel experiment for now, validated against the original outputs.

## Notes
Branched fresh from the restructured `main` (#163) to keep this PR scoped cleanly to just the segmentation speedup work.


annawoodard (changes requested):
Thank you, Sarit!! This looks like exactly the right bottleneck to target, and if the speedup holds, it could make the segmentation step much more usable for the project.

I'd like to check if we should switch to this and then switch it over rather than maintain two parallel paths. Before we switch over, can you add one small before/after validation so we know the fast path is reproducing the existing pipeline?

Please run the old pipeline and the fast pipeline on the same small set of files, ideally ~5-10 files (NOT changing sampling approaches between each, this is just testing the speedup change), and report (mean and standard deviation):

  - old runtime
  - fast runtime
  - output shape match
  - Dice between the old and fast thresholded vessel masks at 0.5
  - max and mean absolute probability difference between old and fast probability maps
  - the exact command/output directory used

Also, can you run `pre-commit run --all-files` and commit the lint/format fixes? One tiny operational fix too: the full-array script currently defaults to the smoke output directory, so please either make the full-run default a separate directory or require `OUTPUT_DIR`.

annawoodard (approved):
Wow!!!! Huge speedup!!! This is great, Sarit. Thank you for your work on this, huge improvement 👏 👏 👏 👏 

I am going to merge this now so that it doesn't hold up anything. But can we put on your todo list for lower priority while other things are running to come back and I would like to remove the old segmentation code so that there is only one path. That will mean some minor changes to this code: once there is only one path, please rename things, e.g. we don't need hte "fast" qualifier anymore, and re-organize so that nothing is redundant, README and docs point to the right places, etc. Please throw that into a github issue so we don't lose it

PR #171, opened 2026-07-02
Implement gnn

spencervenancio:
First implementation or the GNN track.

Adds a data loader that converts each case's centerline `networkx` graph into a `torch_geometric` `data.Data` object. Node features are a configurable list; for now it's populated with peak contrast-arrival time and local vessel radius,
leaving room to extend later.

Training currently uses a single train/val split; k-fold Cv is planned for a future iteration. Validation AUC - 0.5.


annawoodard (changes requested):
This is a great start, @spencervenancio!!

I'm not worried about AUC ~0.5 yet; with voxel nodes and only a couple of features, this is mostly a wiring test. Before adding model complexity, I want us to make the GNN path more auditable and more consistent with the rest of Vanguard.

**1. Please reuse the central split/evaluation code**
We already have shared evaluation machinery:
    - evaluation/build_splits.py
    - evaluation/evaluator.py
    - evaluation/kfold.py
    - evaluation/utils.py
So I don't want the GNN path to grow its own independent split logic around `train_test_split`. For the GNN, make a small dataframe with one row per graph: `case_id`, `pcr`, `dataset`, `site`, `graph_index`, `fold` (if available). Then use the same central split/evaluation tools that tabular and Deep Sets use. The PyG dataset can still hold the graph tensors; this dataframe is just for splits, labels, and reporting. We want to support a predefined fold column so that should be added centrally in `evaluation/build_splits.py`, not only inside `gnn/train.py`, so every model family can use the same fold definition.

**2. Please change the kinetic feature source**
I'm pretty sure you did not mean to compute peak time from the vessel segmentation probabilities. Right now, it looks like peak_time is coming from `run_summary.json["study_files"]`, which points to `*_vessel_segmentation.npz`. That means the current feature is peak vessel-segmentation probability time, not peak DCE enhancement time.

  For the GNN, I want the kinetic node features to come from the raw DCE image series:
`/gpfs/data/karczmar-lab/MAMA-MIAsyn60868042/images/<case_id>/<case_id>_????.nii.gz`

  The basic enhancement curve for a node should be:

```
  curve = dce_4d[:, z, y, x]
  baseline = curve[0]
  enhancement = curve - baseline
```

  Then derive features like:
```
  time_to_peak = argmax(enhancement)
  peak_enhancement = max(enhancement)
  time_to_enhancement = first time enhancement reaches a fraction of peak
  washin_slope = (peak - baseline_or_arrival_value) / time_to_peak
  auc_positive = area under max(enhancement, 0)
```

Please reuse the existing conventions where possible. `features/kinematic.py` already does this kind of calculation, and `_arrival_index_from_enhancement` exists in `graph_extraction/feature_stats.py`. Don't re-invent a separate definition unless there is a reason.

One important detail: the graph/skeleton coordinates are in the same orientation as the saved segmentation/model outputs, not necessarily the raw SimpleITK array orientation. So, before sampling the raw DCE, please verify the alignment carefully. At minimum:

  - load the raw DCE timepoints
  - transform them into the same (z, y, x) space as the support/skeleton mask
  - assert dce_4d.shape[1:] == support_mask.shape
  - make one visual QC overlay for a few cases

Also, do not independently z-score each timepoint before computing enhancement, because that can destroy the kinetic meaning. If normalization is needed, use consistent per-case/reference normalization after computing enhancement.

**3. Add a cache manifest**

The cache is useful, but it can become dangerous if stale graphs are reused after changing features. Please add a cache_manifest.json next to processed/data.pt recording:

  centerline root
  raw DCE images root, if used
  labels path
  label column
  case list or dataset filter
  node mode
  node features
  feature source: raw_dce vs vessel_segmentation
  code commit
  number of graphs
  label counts
  date/time built

When loading a cache, compare the requested settings to the manifest. If they do not match, fail loudly or require an explicit override.

**4. Save split and prediction artifacts**

Every training run should write enough files that we can inspect what happened without reading the code.

  Please save:

  split_manifest.csv
  predictions.csv
  metrics.json
  config_used.json

  split_manifest.csv should include:

  case_id
  graph_index
  dataset
  site
  pcr
  fold or split
  train_or_val
  num_nodes
  num_edges

  predictions.csv should include:

  case_id
  dataset
  site
  pcr
  y_prob
  y_pred
  fold or split

This will let us check whether the model is learning biology, site effects, graph size, or something else. Ideally this would be implemented in the same/similar central place like the splitting. 

 ** 5. Add graph/data QC summaries**

  Please make the dataset build write a graph_qc.csv with one row per graph:

  case_id
  dataset
  pcr
  num_nodes
  num_edges
  num_connected_components
  mean_degree
  missing_feature_count
  nan_feature_count
  feature_min/max/mean/std for each feature

For this GNN, graph size and feature ranges are not side details. They are possible confounders. I specifically want to be able to plot:

  num_nodes vs pcr
  num_nodes vs dataset/site
  prediction vs num_nodes
  feature distributions by dataset/site
  feature distributions by pcr

**  6. Add sanity-check runs before fancier GNNs**

  Before trying a more complex architecture, please add an overfit test. If the model cannot overfit 8-16 cases, we debug the data/model wiring. If it can overfit but validation stays random, then we ask whether the representation and  features have signal.

**  7. Keep voxel-node as a smoke test, but don't over-optimize it**

Voxel-node graphs are okay for proving the pipeline works, but I think they are probably too large for this dataset size. The next scientific representation should probably be segment-as-node, where each vessel segment has features like:

  length
  radius
  tortuosity
  distance to tumor
  tumor shell
  time to enhancement
  time to peak
  peak enhancement
  washin slope
  AUC

For now, though, the goal is not better AUC. The goal is to prove that the dataset, enhancement features, labels, splits, cache, and training loop are doing exactly what we think they are doing.

annawoodard:
also resolve conflicts with main please

spencervenancio:
1. **Please reuse the central split/evaluation code**
Done. Cross validation through the shared `evaluation/` package is now implemented
2. **Please change the kinetic feature source**
Done. Bug with kinetic feature source resolved to use `images/`
3. **Add a cache manifest** 
Extended the unit tests to cover cache invalidation on the cases field (previously only `node_features` was tested), then ran 4-step Slurm sequence against a disposable cache dir: 
- build identical rerun (confirmed genuine cache hit via unchanged file mtimes)
- tweaked cases param without override (confirmed a loud RuntimeError blocking reuse, cache left untouched) 
- same tweak with --force-rebuild (confirmed a genuine, content-verified rebuild with a fresh manifest 
- archived old cache
4. **Save split and prediction artifacts**
Done. Now implemented.
5. **Add graph/data QC summaries**

Screenshot 2026-07-07 at 4 47 25 PM

Screenshot 2026-07-07 at 4 47 49 PM

6. **Add sanity-check runs before fancier GNNs**

image

7. **Keep voxel-node as a smoke test, but don't over-optimize it**
Sounds good, I will keep this in mind.

annawoodard (approved):
This looks great, Spencer!! Minor tweak: I think you brought back `CLAUDE.md` in `.gitignore`, please remove that. Also, I'm confused why your loss goes up for fold 1 in your overfit test. Let's make sure we understand that. 

I'm assuming you'll merge when ready, everything else looks great!

PR #173, opened 2026-07-06
Restore vanguard-blood-vessel-segmentation as a pinned git submodule

SBose-12:
Reverts an earlier regression where the submodule became an untracked/ gitignored bare clone, even though the README already documented `git clone --recursive` for it. Pinned to f8b0a3a.


annawoodard (approved):
thank you @SBose-12!!!!

PR #174, opened 2026-07-07
Chore: retire deprecated /net/projects2/vanguard/ paths (Step 0 of modularization)

SBose-12:
## What this does

Step 0 of the multi-dataset modularization plan (see `Modularization-design-doc.md`
§3, §9). Mechanical path cleanup, done independently and ahead of any adapter/
architecture changes so it doesn't get entangled with that work.

No behavior change intended for the canonical use case. Several of the fixes
below *do* change behavior in the sense that they make previously-broken
configs actually work for the team (see below).

## Changes

**1. Canonical MAMA-MIA root swap**
Replaces `/net/projects2/vanguard/MAMA-MIA-syn60868042/` with the confirmed
canonical root `/gpfs/data/karczmar-lab/MAMA-MIA-syn60868042/` across 39 files
(code defaults, run configs, docs). Pure prefix swap, verified line-by-line.

**2. `pcr_labels.csv` provisioned at the canonical root**
This file did not exist at the new root even though 27 configs reference it
there. Resolved by copying a verified working copy (previously only in a
personal workspace) to the canonical root:
- md5-matched against the original
- cross-validated against source `patient_info_files/*.json`: 1491 labeled
  rows, 15 null-pcr cases correctly excluded, 6/6 spot-checked case_ids match
- original workspace copy left untouched

**3. Fixed three config paths pointing at a private, unreadable home directory**
`ispy2_smoke.yaml` and `deepsets_ispy2_smoke.yaml` referenced
`/ess/home/home1/t-9svena/...` (mode `drwx------`) — these smoke-test configs
only ever worked for that one user:
- `labels_csv` in both → repointed to the in-repo `configs/ispy2_smoke_labels.csv`
  (a distinct, purpose-built 8-row smoke sample, not the full labels file —
  confirmed this is the correct target, not just a naming coincidence)
- `deepsets_manifest_csv` → repointed to a repo-relative `experiments/...` path.
  Also corrected a stale comment claiming this path is overwritten by the
  runner at run time — traced the code and confirmed `train.py` reads this
  config value directly, so the old path wasn't cosmetic; the smoke test was
  genuinely broken for anyone but one person. It now resolves correctly for
  the whole team.

## Deliberately out of scope

- `centerlines_tc4d/studies`, `vessel_segmentations` — derived/per-user output
  locations, no single canonical home yet; better handled once output dirs
  move into run config (later migration steps).

## Follow-up filed separately (not blocking this PR)

- `radiomics/scripts/make_labels_from_json.py` silently skips JSONs keyed by
  `patient_id` (script expects `case_id`), so it currently cannot regenerate
  `pcr_labels.csv` from source JSONs. Doesn't block this PR since the labels
  file was provisioned directly and validated, but should be fixed if the CSV
  is ever regenerated.

## Validation

- `git diff` reviewed line-by-line — pure prefix/path swaps only, no logic
  changes.
- `pre-commit run --all-files` passes.
- `pcr_labels.csv`: 1491 rows validated against source JSONs.
- Smoke configs verified to parse and resolve to real files after the fix.

Refs: `Modularization-design-doc.md` §3, §9 (Step 0)


annawoodard (changes requested):
Great work, Sarit!!!! 🎉 🎉 🎉 

I really appreciate that you kept this separate from the adapter/scaffolding work. That makes the diff much easier to review and keeps the mechanical path cleanup from getting tangled with design decisions. The PR description is also excellent: it explains what changed, what was validated, and what you intentionally left out of scope.

I agree with the overall direction here. One thing I'd like you to tighten before merge is how we handle the remaining MAMA-MIA root paths. A few /net/projects2 MAMA-MIA paths still remain in runnable scripts/notebooks:

  - `segmentation/slurm/submit_batch_segmentation_array.sh`
  - `segmentation/slurm/submit_batch_segmentation_array.slurm`
  - `radiomics/slurm/slurm_generate_kinetic_maps.sh`
  - `analysis/clinical_imaging_exploration.ipynb`

I agree that derived output roots like `centerlines_tc4d` and vessel_segmentations can wait for a later output-dir cleanup. But for MAMA-MIA images/masks/patient info, I'd rather avoid replacing one hardcoded absolute root with another hardcoded absolute root everywhere. /gpfs/... is the current correct location for us, but it is still environment-specific in the same basic way /net/... was. Sorry I missed this detail in the plan.

For the shell scripts, could you use a single overridable root instead? Something like:

```
  MAMAMIA_ROOT="${MAMAMIA_ROOT:-/gpfs/data/karczmar-lab/MAMA-MIA-syn60868042}"
  IMAGES_DIR="${IMAGES_DIR:-${MAMAMIA_ROOT}/images}"
  MASKS_DIR="${MASKS_DIR:-${MAMAMIA_ROOT}/segmentations/expert}"
```

That keeps this PR lightweight, but avoids spreading another arbitrary absolute path through scripts. I don't think you need to solve full config-derived path composition in this PR; that probably belongs with the real adapter/config refactor. This smaller root-variable pattern feels like the right middle ground for Step 0.

Can you also please merge main into this branch so this can be automatically merged

annawoodard (approved):
Looks great!!  Thank you, @SBose-12! 🎉

PR #175, opened 2026-07-07
feat: scaffold datasets module and adapter contracts (Step 1 - modularization)

SBose-12:
What this does
Step 1 of the multi-dataset modularization plan (see Modularization-design-doc.md §9). This PR introduces the foundational, purely additive scaffolding for the new datasets/ module.

Crucially, this code is wired to nothing—no existing pipeline stages call or consume these modules yet, making this a zero-risk, isolated architectural layout.

Branch Hygiene
Base: Branched cleanly off origin/main.

Dependencies: Verified that all Step 0 commits (chore/retire-net-projects2-mamamia-paths) are absent. This PR has zero dependency on the unmerged Step 0 work.

Changes
1. Core datasets/ Module Scaffolding

DatasetAdapter: Implemented the base class contract outlining required methods per §4 of the design doc (including discover_cases(), case_dataset_name(), group_key(), load_timepoints(), preprocess(), load_clinical(), load_labels(), etc.). The constructor accepts an injected root: Path rather than hardcoding paths (§10, Decision 5).

MamaMiaDataset(DatasetAdapter): Set up the adapter implementing canonical MAMA-MIA behavior. Per §6.5, this requires zero method overrides, pulling existing default configurations and structural rules from the legacy setup into base-class parameters.

UChicagoDataset (uchicago.py): overrides the six manifest/identity/label methods identified in §6.5 (discover_cases, load_timepoints, preprocess, case_dataset_name, group_key, load_labels), plus two class-attribute defaults (default_split_policy, report_by).

Factory Function: Added a lightweight factory utility to resolve and instantiate the correct adapter based on the run config.

2. Configuration Knobs (config.py)

Appended two new fields to DEFAULT_CONFIG to govern dataset routing per §9:

dataset_root: Directs the adapter to its target directory.

split_policy: Standardizes data splitting strategies (auto / compute / provided).

3. Isolated Testing

Added targeted unit tests exercising adapter construction, ensuring MamaMiaDataset functions cleanly with zero overrides, and validating that the dataset factory instantiates classes correctly based on mock configurations.

Validation
Local unit tests for the new datasets/ module pass successfully.

Verified no active references or pipeline wiring hook into the new module.

pre-commit run --all-files passes cleanly.


annawoodard (changes requested):
Sarit, fabulous work!!!! 🎉 🥳 


Overall this makes sense: the adapter contract is readable, the MAMA-MIA behavior is documented clearly, the UChicago adapter fails loudly where preprocessing is not implemented yet, and the factory gives us a simple seam for later wiring. That is exactly the kind of incremental architecture PR I want here: additive, isolated, and easy to reason about.

The main design thing I want us to settle now is the package name. I'm hesitant about a top-level package called datasets because it collides with the common Hugging Face datasets package. Even if we are not using that package today, import-shadowing bugs are annoying and hard to diagnose later.

I also don't want something redundant like `vanguard_datasets`. Shorter options I'd consider:

  - `cohorts`: my favorite; short and domain-specific
  - `sources`
  - `adapters`

My vote is cohorts. These classes are not just generic datasets; they encode cohort-specific identity, labels, splits, layout, and preprocessing assumptions. `from cohorts import MamaMiaDataset` reads naturally and avoids the external package collision.

On tests: I appreciate that you added them, but I want us to be disciplined here. Tests are only worth carrying if they run automatically. Right now, our GitHub workflow sets up the environment and runs pre-commit, but it does not run pytest. That means new tests can silently go stale, which becomes a maintenance burden rather than safety.

If we keep tests in this PR, please add a CI step after pre-commit:

```
  - name: Run unit tests
    shell: bash -l {0}
    run: python -m pytest tests
```

 I'd also trim the adapter tests to the stable contract rather than the exact implementation shape. The useful tests include factory selection, invalid config errors, split-policy resolution, and "UChicago preprocessing should not silently run until implemented". I'd remove or soften tests that assert MAMA-MIA has zero overrides forever or that UChicago has exactly a specific override set. Those freeze the current scaffold too tightly while the design is still evolving.

Small consistency note: the PR description says UChicago has five explicit overrides, but the test expects six. The code seems reasonable; I'd just align the wording so the PR description, tests, and design notes teach the same contract.

Finally, please merge main into this branch

annawoodard (approved):
Lookls great! I like the `cohorts` choice. I also appreciate that this is a true scaffold that defines the adapter contract and factory without changing the current behavior. 

I like the tests much better now. They check useful behavior, like adapter construction, split policy resolution, MAMA-MIA orientation behavior, and the fact that UChicago preprocessing is intentionally not implemented yet, without locking us into brittle implementation details. Adding pytest to CI also makes the tests meaningful rather than just local documentation.

One small non-blocking cleanup: a few comments/docstrings reference docs/modularization-design.md, but I don't see that file in the repo. Could you reword those references so future readers are not pointed to a missing file?

Overall: great job!!! 👏 👏 👏 👏 

I'm approving, please fix the docstring refs and merge

PR #177, opened 2026-07-08
Consolidate vessel segmentation into one validated pipeline

SBose-12:
Folds the fast, in-process batched segmentation pipeline (previously living separately in faster-segmentation-test/, validated at 16.15x speedup with Dice 0.9998-1.0 vs the old pipeline) into segmentation/ as the sole implementation. Removes the old subprocess-per-file pipeline it replaced (run_vessel_segmentation, process_single_file, the old CLI) now that duplication between the two locations is gone.

Also bumps segmentation/slurm/submit_batch_segmentation_array.slurm's SBATCH resources (gpuq, 4 CPUs, 32G) to match what the batched pipeline actually needs, since the array job still had the old serial pipeline's minimal resource request.

validate_speed_and_accuracy.py is kept but frozen/historical: it drove the old pipeline's --file-index CLI, which no longer exists, so it can't actually re-run a comparison -- it documents how validation_results.md's numbers were produced. Re-ran the CPU-safe equivalence tests (test_batching_equiv.py, test_preprocess_parallel.py) against the consolidated code; both still pass bit-identical.


annawoodard (approved):
Great job, Sarit!!

Folding the validated fast path back into segmentation/ and deleting the duplicate makes the repo easier to reason about. The PR description is also helpful because it clearly separates the validated fast pipeline from the now-historical validation scripts.

One thing I'd like fixed before merge: the Slurm scripts currently use `SLURM_SUBMIT_DIR` as `PROJECT_ROOT`, but the README recommends launching from inside `segmentation/slurm`. In that usage, `SLURM_SUBMIT_DIR` is `.../segmentation/slurm`, not the repo root, so paths like `${PROJECT_ROOT}/segmentation/batch_segmentation.py` will resolve
  incorrectly.

Could you make `PROJECT_ROOT` explicit and stable? For example, the wrapper can pass PROJECT_ROOT="${PROJECT_ROOT}" into sbatch, and the .slurm scripts can use that env var instead of deriving from `SLURM_SUBMIT_DIR`. The important thing is that following the documented entry point should resolve the repo root correctly.

I trust you to make this change, so please fix this and then merge without another iteration from me. I've approved it

annawoodard:
Nice naming also :)

PR #179, opened 2026-07-08
Duke Symmetry Lines

aakrithiram:
- Add a self-contained DUKE final symmetry-line workflow using the inner-wall method, which measures the distance from the center of the image until it hits the inner side of the mask. That distance is then divided by 2, and the symmetry line is placed there. 
- Add the final DUKE symmetry-line CSV for 284 cases
- Add a summary generator, QC documentation, and focused tests


annawoodard (changes requested):
This is a great contribution!!

The symmetry-line question is easy to handle informally case by case, but much harder to make reproducible, so I love this PR turns the Duke side-assignment workflow into an explicit script, final CSV, summary, and tests. I also like that the final export is intentionally constrained to inner_walls rather than mixing in fallback methods. The algorithm for finding the centerline is beautiful :)

The tests here feel helpful: they check the core midpoint behavior without overtesting every implementation detail. 

One thing to tighten before we merge: The README says to regenerate from `tabular/mama_mia_bilateral_unilateral_cancer_manifest.csv`, but that manifest is not in the PR, so the command will fail from a clean checkout. Could you either commit the manifest used for this export, or update the script/README so the user must pass `--manifest` pointing to the shared canonical location?

Overall, great work and a solid foundation for the symmetry model!

annawoodard (approved):
Updates look great!

As discussed on slack next time you'll add them as new commits

PR #180, opened 2026-07-09
Feat/adapter in feature extraction(Step 2 of Modularization)

SBose-12:
Routes cohort identity through the Step 1 `DatasetAdapter`: the `dataset` column
now comes from `adapter.case_dataset_name(case_id)` instead of the directory
layout. Stacked on Step 1 — base this PR on `feat/dataset-adapter-scaffold`.

**How it works.** `tabular/cohort.py`'s feature build takes an optional adapter
(`None` = today's behavior, exactly). Both entry points (`tabular/train.py`,
`modeling/ablation.py`) build it via the Step 1 factory, which returns `None` for
every config without a `dataset:` block — so no current run changes.

**Validation.** `scripts/validate_adapter_feature_parity.py` proves
`features_engineered_labeled.csv` is byte-identical with/without the adapter on
`all_datasets.yaml`: PASS, `(1491, 1286)`. Plus CI-safe parity/live-seam unit
tests. `ruff` clean; suite 100 passed / 2 skipped.

**Docs.** Adds `cohorts/README.md` and repoints code comments to it (full design
doc kept out of the repo). Touches a few Step 1 `cohorts/*.py` files (docs-only).

**For review.** `cohort` is vestigial at this stage — the combined table uses
cohort-independent `case_dataset_name` (noted at the call site).


annawoodard (approved):
This looks GREAT, Sarit!

I like how conservative this Step 2 is: the adapter is threaded through the tabular feature build, but configs without a `dataset: block` still follow the existing behavior. The parity test is also exactly the right kind of test here: it proves MAMA-MIA output is unchanged when directory identity and case-id identity agree, and also proves the new seam is live when they disagree.

I ran the tests locally, including a temporary merge with current main, and they pass. This is a very solid incremental migration step.

Please merge with main and merge after you merge #175.

PR #181, opened 2026-07-10
Feat/uchicago adapter wiring(Step 3/5 - Modularization)

SBose-12:
Makes UChicagoDataset usable against the real internal ultrafast manifest (181 labeled exams), scoped to the manifest-driven path only — no vessel pipeline, since no UChicago segmentations/centerlines exist yet and imaging stages aren't adapter-wired until Step 4.

What's included:
- preprocess() is a documented pass-through — manifest images are already preprocessed upstream (policy_name = hfdp_t1_v1); does not fall back to the MAMA-MIA orientation transform.
- load_folds() / resolve_folds() surface UChicago's shipped patient-grouped CV folds, and — importantly — are now actually consumed: tabular/train.py merges them onto the feature table, and a run opts in via model_params.split_mode: predefined (configs/uchicago.yaml). Previously this was parsed/validated but wired to nothing.
- load_timepoints() rebases phase-file paths against the injected root, so a moved manifest directory doesn't leave stale paths.
- configs/uchicago.yaml selects the dataset via the factory (canonical /gpfs manifest root).
- Validation: scripts/validate_uchicago_manifest.py checks the adapter against the real manifest (discovery, labels, folds, sub-sources, timepoints) — all passing. Full test suite: 114 passed / 2 skipped.

Known open item: the preprocess() pass-through assumes UChicago's hfdp_t1_v1 volumes are already correctly oriented for the downstream vessel/graph pipeline. That assumption isn't verified end-to-end (no UChicago imaging flows exist yet). Flagged in cohorts/README.md.

### DIFF
--- a/tabular/train.py
+++ b/tabular/train.py
@@ -15,7 +15,8 @@ from typing import Any
 
 import pandas as pd
 
-from cohorts.factory import build_adapter_from_config
+from cohorts.base import DatasetAdapter
+from cohorts.factory import build_adapter_from_config, resolve_folds
 from evaluation import FoldResults
 from evaluation.build_splits import create_splits_for_dataframe
 from evaluation.kfold import FoldSplit
@@ -135,6 +136,9 @@ def prepare_evaluation_context(
     stratum_col = model_params.stratum_col
     if stratum_col:
         drop_cols.add(str(stratum_col))
+    # Eval bookkeeping, not a feature -- must not leak into X even when unset
+    # (harmless to drop a column that isn't present).
+    drop_cols.add(str(model_params.split_col))
 
     X = df.drop(columns=[c for c in drop_cols if c in df.columns], errors="ignore")
     if X.empty:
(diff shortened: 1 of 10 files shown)


tabular/train.py
```diff
@@ -283,6 +287,55 @@ def run_cross_validation_from_context(
     return fold_results_list, nested_rows
 
 
+#: Cap on how many case ids to list in the missing-fold error before truncating.
+_MAX_MISSING_CASES_SHOWN = 10
+
+
+def _apply_provided_folds(
```
annawoodard, line 333: Could we make this merge fail closed by validating unique case IDs, using validate="one_to_one", and requiring every modeled row to receive a non-null fold? Right now duplicates can leak a case across train/validation, while NaN creates a split with zero validation rows rather than the documented loud failure. Please also reject collisions with case_id and the configured label column.

annawoodard (changes requested):
This looks great!!

There's a lot to like here: the adapter is exercised against the real 181-exam manifest, all 3,218 phase paths resolve, the provided folds preserve patient grouping, and excluding the fold column from model inputs is exactly right. The focused tests, path rebasing, and explicit orientation caveat are also strong.

 I have three contained requests before merge:

 1. Please refresh this branch from current main. 

 2. Please make provided-fold attachment fail closed: require exactly one non-null fold per modeled case and reject unsafe split_col collisions. A missing fold currently produces an empty validation split, while duplicate mappings can put the same case in both training and validation.

 3. The documented split_policy: compute option needs to preserve patient grouping. UChicagoDataset.group_key() is not currently consumed, so recomputed folds use exam-level CV. The real manifest has 181 exams from 143 patients, including 36 patients with multiple exams.

Besides those small tweaks this looks good to go!!

tabular/train.py
SBose-12: Pushed fixes for all three: fail-closed fold attach (validate="one_to_one" + completeness + collision checks), patient-grouped compute splits via group_key, and confirmed the branch is already on current main.

annawoodard (approved):
Re-reviewed the current head. The requested fail-closed fold validation and patient-grouped computed-split wiring are addressed; focused tests and the real-manifest validation pass. No blocking issues from me.

annawoodard (approved):
All the feedback fixes look great, thank you!!! 👏

PR #183, opened 2026-07-10
Feat/implement new graphs

spencervenancio:
**WIP**: GNN graph representations - voxel vs segment vs junction
--- 
Adding two new graph representations for the vessel GNN (previously voxel-only) plus a matched comparison experiment. 

**What's in here**:
- Segment mode (`segment_graph.py`) - each vessel segment is a node instead of each voxel
- Junction mode (`junction_graph.py` + `EdgeGNNClassifier`) - junctions are nodes, segments are edges with edge features
- Wired both through `data_loader.py`, `train.py`, configs, and build scripts alongside the original voxel mode
- Matched 3-way comparison (`make_fold_assignments.py`, `gnn_*_compare.yaml`, `gnn_graph_compare_plot.py`) with the same cohort/folds/features across all 3.
- TTE NaN sentinel handling documented in `AUDITING_RESULTS. md`

**Result so far**: matched 5-fold CV, ~chance AUC for all three (voxel 0.526, junction 0.511, segment 0.502), which is unsurprising since it's only 6 basic features with no tuning. Point was to prove the comparison harness is fair; it is. Full details in `experiments/graph_repr_compare_v1/README.md`.

**Not done**: edge-feature NaN QC, pure-cycle handling, gpfs quota still full so runs went to scratch. v2 with richer features is next.


annawoodard:
image

I'm adding the training curves you shared in Slack here for posterity

Nice work getting all three graph representations implemented and running!!!!

Looking at the learning curves, train AUC consistently rises to roughly 0.57-0.58, while val AUC is flat or declines and varies substantially by fold. This looks like a weak training-set pattern that does not generalize consistently, rather than simply insufficient training. I would not expect more epochs alone to fix it.

One potentially important source of fold heterogeneity is the mixture of bilateral and unilateral acquisitions. If bilateral graphs contain both breasts, the pCR label pertains to the tumor-bearing side, while global mean pooling averages vessels from both sides. That could dilute the tumor-local signal and create a substantial coverage/domain shift. Laterality is not strongly associated with pCR by itself in the existing audit, but it is associated with large graph-feature differences and is confounded with the dataset.

Could you ask Aakrithi for her breast-splitting algorithm? I think it would be worthwhile to build a harmonized single-breast dataset:

  - Leave native unilateral cases unchanged.
  - For bilateral cases, generate left/right breast masks and retain only the tumor-bearing side, selected by overlap with the tumor mask.
  - Apply that mask to the skeleton/support graph while retaining the original DCE coordinate system.
  - Do not create two independently labeled samples from one patient.
  - Exclude or separately handle cases with unknown tumor side, tumor overlap on both sides, or genuinely bilateral disease.

Before running the full cohort, validate the splitter visually on approximately 20-30 bilateral cases across datasets. Overlay the breast mask, tumor mask, skeleton, and retained graph; verify that the tumor-bearing breast is always selected, that there is no contralateral or midline leakage, and that graph-size distributions become comparable to native unilateral cases.

For the minimal experiment, I would start with only the voxel model and compare:

  1. Current mixed full-exam graphs.
  2. The same cases and frozen folds, but with every case represented by one breast.
  3. A native-unilateral-only reference cohort.

Use identical features and hyperparameters, assert that arms 1 and 2 retain exactly the same cases, and run three to five seeds. If the single-breast arm improves consistently, then rerun the segment and junction representations. If it does not, there is no reason to pay for the full representation matrix yet.

  For evaluation, please also:

  - Show dataset -> laterality -> pCR counts by fold. If materially imbalanced, regenerate and freeze folds balanced on these variables where sample counts permit.
  - Plot pooled out-of-fold AUC across epochs for diagnosis, alongside the individual fold curves.
  - Report OOF AUC by dataset and laterality, especially within ISPY2.
  - Fit a simple logistic regression or XGBoost baseline on summaries of the same six features: mean, SD, q10/q50/q90, TTE no-arrival fraction, and the already-recorded node/edge counts.

  The outcomes would distinguish the likely failure modes:

  - Single-breast input improves both models: bilateral coverage was diluting signal.
  - Tabular summaries beat the GNN: pooling or message passing is the bottleneck.
  - Both models remain near chance: the six inputs lack generalizable pCR signal.
  - The same folds remain good/bad across seeds: cohort or acquisition differences are driving the result.
  - Fold behavior changes across seeds: the apparent representation ranking is mostly optimization variance.

If the GNN readout is the bottleneck, mean+max pooling with standardized graph size/density features would be a reasonable next ablation.   If the inputs are the bottleneck, I would prioritize within-patient intensity normalization and tumor-relative vascular information. MRI intensities are scanner-dependent, so compute relative enhancement as ((S(t)-S_0)/(S_0+\epsilon)), using the pre-contrast phase as (S_0), rather than relying on raw intensity differences. If a consistent normal-tissue region can be defined, such as normal parenchyma away from the tumor, its median enhancement curve could provide a patient-specific reference for vessel-to-reference peak, AUC, or arrival-phase differences. Because the preprocessed MAMA-MIA files do not appear to retain elapsed acquisition times, temporal features should be described explicitly in phase units, not seconds, and should favor normalized peak/arrival phase, early-to-late enhancement ratios, and curve-shape or washout descriptors. Results should also be checked within each source dataset because their phase counts and acquisition protocols differ. Finally, use the tumor mask to attach signed tumor distance and distance-shell membership to vessel nodes or edges, then compute vascular density, length, radius, tortuosity, branching, and enhancement summaries inside and near the tumor. This would allow the model to focus on biologically relevant peritumoral vessels rather than averaging across the entire breast network. True time-to-enhancement and slopes per second can be revisted when we switch to the UChicago data, where we have the raw DICOMs readily available.

 Finally, the frozen-fold check confirms that cohort assignments match, but it does not isolate representation as the only difference, since the junction arm also changes its architecture and feature carriers. I would describe this as a comparison of representation-plus-architecture pipelines and characterize the current result as _these six globally pooled features have not yet demonstrated a generalizable pCR signal_.

spencervenancio:
## Harmonized Single-Breast Dataset
Using @aakrithiram's breast-splitting algorithm, we can build a single-breast dataset. To peform QC on the splits, I inspected 30 cases. The splits were verified to be working correctly, and we can summarize the datasets by a few examples 

image
image


A couple of things worth flagging from the QC pass itself, not just the result:
- Caught a real bug along the way: the tumor segmentation files had an axis-orientation ambiguity that was silently sending the splitter to the wrong side. First pass excluded 77% of cases as "tumor outside breast mask" — root-caused to the flip, fixed, re-verified on 11 real cases, dropped to 0% false-exclusions.
- The full-cohort run surfaced that **89% of ISPY2's bilateral cases (153/172) are missing tumor masks entirely** — not visible in the 30-case QC sample since case selection there already required all inputs present. Net effect: the harmonized cohort is 94% DUKE (289/308 retained bilateral cases), more skewed than the raw bilateral cohort was.

## Minimal Experiment (<u>Null Result</u>)

We can summarize the results of the minimal experiment by the following table:

| arm | n cases | mean AUC (across 3 seeds x 5 folds) | AUC std |
|---|---|---|---|
| 1: mixed baseline | 1332 | 0.5090 | 0.0352 |
| 2: single breast | 1332 (identical case set to arm 1) | 0.5076 | 0.0353 |
| 3: unilateral only (ceiling, different/smaller cohort) | 1035 | 0.5161 | 0.0388 |

Arm 1/2's case-set identity is asserted, not assumed — `analysis.assert_arm_parity` diffs the two caches' actual built cohorts and confirms the same 1332 case IDs. Arm 3's smaller cohort isn't just smaller, it's a different mix: since bilateral cases skew heavily DUKE (see above), unilateral-only skews the other way, toward ISPY1/ISPY2/NACT — so it's a ceiling/sanity-check, not a case-matched comparison.

Paired arm2-vs-arm1 delta (by seed x fold, n=15) is positive in only 6/15 pairs, mean −0.0014 — no consistent direction, and fold-to-fold variance is unchanged (0.0377 vs 0.0378 std within-seed). Per plan, since single-breast didn't help consistently, **we are not re-running the segment/junction representations on this hypothesis.**

> **Interpretation:** Due to the simplicity of our set-up, and the results of the experiment, we are not currently concerned about any imbalance cause by unilateral vs. bilateral aquisitions 
>
>Because both arms sit at or below chance (~0.51 AUC) there's this little signal to begin with. The null result to our experiment is less informative than it would be if the mixed baseline had something real to dilute
>
> This is a *floor-effect* issue. That is we can't fully distinguish "dilution isn't a problem" from "there's barely any signal here regardless of how it's pooled." 
>
> So *not a concern at present* <u>not</u> ***never** a concern.* We can revisit this at a later date, when the baseline achieves more predictive performance

## Further

### Fold Composition

DUKE is 100% bilateral and ISPY1/NACT are 100% unilateral in this cohort — laterality is basically a proxy for dataset here. ISPY2 is the only dataset with a real mix, and even there bilateral is tiny (19 cases total across all 5 folds, sometimes zero pCR-positives in a fold). Fold-level pCR rate swings a fair amount within datasets though — NACT 0.09–0.25, ISPY1 0.18–0.48 across the 5 folds.

> **Interpretation:** rebalancing folds by laterality isn't really actionable — ISPY2's bilateral n is too thin to stratify further. Rebalancing by *dataset* looks more promising given the pCR-rate swings above, but that changes the frozen split everything else here depends on, so <u>treating as an open decision, not doing it silently</u>.

### Out-of-Fold AUC by Dataset / Laterality

| dataset | mixed baseline | single breast | unilateral only |
|---|---|---|---|
| DUKE | 0.489 | 0.488 | — |
| ISPY1 | 0.437 | 0.437 | 0.434 |
| ISPY2 | 0.481 | 0.481 | 0.489 |
| NACT | 0.532 | 0.528 | 0.513 |

Every dataset sits at or below chance in both arms — ISPY1 is actually *below* 0.5. Within ISPY2, bilateral-vs-unilateral AUC swings a lot between arms (0.54 → 0.47), but that's pooled over 19 real cases (57 rows across 3 seeds) — not something to read much into on its own.

### Fold Stability Across Seeds

This is the more useful one. Fold ranking is nearly identical across all 3 seeds, in both arms (Spearman rank correlation 0.9–1.0 on fold AUC): fold 0 is *always* the worst (~0.45–0.46), fold 2 is *always* the best (~0.55–0.56), regardless of seed.

> **Interpretation:** the val-AUC instability looks structural — tied to which cases land in which fold (cohort/acquisition composition) — <u>not optimization noise that more seeds or epochs would average out</u>.

### Tabular Baseline

Fit logistic regression and XGBoost on hand-summarized versions of the same 6 features (mean/std/q10/q50/q90 per feature, TTE no-arrival fraction, node/edge counts — 33 columns), same cohort and folds as arm 1. No new graph build, reused the existing cache.

| model | mean AUC | std |
|---|---|---|
| logistic regression | 0.499 | 0.042 |
| XGBoost | 0.540 | 0.018 |
| *(reference)* GNN arm 1 | 0.509 | 0.035 |

Caveat on the std column: both tabular models turned out fully deterministic at these settings (identical AUC across all 3 "seeds"), so their std is *pure* fold variance — not directly comparable to the GNN's, which mixes fold variance with real training-run noise. The apples-to-apples comparison is against the GNN's isolated fold-only std (0.0377) — XGBoost's fold-to-fold spread is still about half that.

More interesting than the aggregate gap: **per-fold AUC doesn't track between models.** The GNN's worst fold (0) is XGBoost's *best* fold, and XGBoost's worst fold (4) is middling for the GNN. That complicates a pure "cohort/acquisition alone explains it" story — a real confound should hurt every model on the same fold; here it doesn't.

> **Interpretation:** XGBoost's edge (+0.03 AUC, half the fold variance, and specifically stronger on the GNN's weakest fold) is directionally consistent with pooling/message-passing leaving some signal on the table — but it's an untuned first-pass baseline on only arm 1's cohort, and 0.54 still isn't a clinically meaningful AUC on its own. <u>Real, useful data point — not yet a verdict on the architecture.</u>

### Open Items

- Fold rebalancing by dataset — decision not made yet (see above).
- Pooled OOF-AUC-by-epoch plot — not built; needs per-epoch per-case prediction logging in `gnn/train.py` (currently only saves an aggregate AUC per fold per epoch, not the per-case probabilities needed to pool across folds at a given epoch).
- Extending the tabular baseline to arms 2/3, or tuning it, before leaning on the XGBoost-vs-GNN gap.
- The larger feature-engineering agenda (relative-enhancement normalization, tumor-relative reference region, phase-unit timing, tumor-distance/peritumoral vessel features) — not started, explicitly future work, some of it pinned to the UChicago data.

annawoodard (approved):
Spencer, you've done an incredible amount of careful work here. Great job!!! I especially appreciate that you didn't just run the requested comparison: you found the tumor-mask axis issue, manually verified the correction, kept the experimental arms exactly matched, investigated the dataset and fold composition, and followed the planned stopping rule when single-breast processing didn't improve performance. Your interpretation is appropriately cautious, especially around the floor effect and the modest XGBoost result. I think most of the scientific follow-up is complete. Before we merge, there are three integration issues to address. 

1. First, please rebase onto the current main branch and make sure the voxel, segment, and junction graphs all use the same UFAST-aware kinetic policy introduced in PR #188: baseline-frame averaging, relative enhancement, physical timestamps, and the HR_UFAST alignment checks. Please add a small cross-mode regression test using a multi-frame baseline and irregular timestamps.
2. Second, please fit the clinical and morphometry preprocessing separately within each training fold. Right now, the imputation and categorical encoding are fitted using the entire cohort before cross-validation. This means that validation cases contribute to the means or modes used to fill missing training values and to the set of categories learned by the encoder. The outcome labels aren't being used, so this isn't direct label leakage, but the training representation still contains information about the validation distribution that wouldn't be available for a genuinely unseen cohort. That can make cross-validation results slightly optimistic and, more importantly, makes the evaluation procedure inconsistent with how the model would be deployed. For each fold, please fit the imputer and encoder on the training cases only, then use those fitted transformations on the validation cases, treating previously unseen categories as unknown.
3. Third, for the tabular baseline, please calculate TTE summary statistics only over nodes with observed arrival and keep the no-arrival fraction as a separate missingness feature. Including the -1 sentinel in the mean and quantiles makes those features a mixture of actual arrival timing and the prevalence of missing arrivals, even though missingness is already represented explicitly.

This is a very substantial contribution!! You've addressed the scientific feedback with a level of care and thoroughness that I really appreciate.

I am assuming you'll merge when these items have been addressed so I'm approving now

spencervenancio:
1. Resolved as of merge commit [f6543c4](https://github.com/dsi-clinic/vanguard/pull/183/commits/f6543c4ff3cdf905cecf40868455b073d242eb00). Voxel/segment/junction all now go through `node_kinetic_features` with a cross-mode regression test. **All tests passing**.
2. Fixed in [4ee1340](https://github.com/dsi-clinic/vanguard/pull/183/commits/4ee134059c1e7478c816a50d2d77e82813b3bd1c). Clinical/morphometry preprocessing now fits per fold on the training split only (was fit once over the whole cohort at build time). Build caches the raw inputs in a sidecar; `gnn/train.py` fits the imputer/encoder on each fold's train cases and transforms val with unseen categories all-zero. Regression test in `tests/test_gnn_graph_feature_folds.py`; existing caches regenerate via `gnn.build_dataset --regenerate-graph-feature-inputs`.
3. Fixed in [9bfe91c](https://github.com/dsi-clinic/vanguard/pull/183/commits/9bfe91c51a52197c1d5a931c41d9247c94a7b30f). `summarize_graph` now computes the `time_to_enhancement` mean/std/quantiles over observed-arrival nodes only (the -1 sentinel is masked out); `tte_no_arrival_fraction` still carries the missingness. Edge case worth flagging: 4/1332 cases have no arrivals anywhere, so their TTE timing summaries are now `NaN` both model pipelines gained a per-fold mean-imputer (fit on train only, same as item 2) to handle it. Test in `tests/test_tabular_gnn_feature_baseline.py`. The tabular_baseline_v1 results predate this and need regenerating (flagged in that dir's README).

PR #184, opened 2026-07-13
Batch effect analysis

SBose-12:
Summary
- Consolidates the ISPY2 batch-effect investigation (5 analysis scripts, 6 sbatch launchers, write-up) into one clean set of commits on main, organized under scripts/batch_effect/.
- Copies all input CSVs to shared, group-readable storage (/gpfs/data/karczmar-lab/vanguard/batch_effect_inputs/, md5-verified) so the analysis is reproducible off personal/scratch paths.
- Fixes two reproducibility bugs found before handoff: sbatch launchers hardcoded a personal home-directory python path, and the PCA script's usage docstring pointed at a personal workspace path — both now use the repo's shared conventions.

Key finding: a strong, scanner-driven batch effect exists in ISPY2 vessel features (AUC up to 0.92 predicting scanner from features alone), but it's orthogonal to pCR outcome, so it's nuisance variance rather than a leakage shortcut. Full write-up in docs/resampling_batch_effects.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -64,3 +64,10 @@ Derived outputs should go in clearly named output folders with enough provenance
 If `agents.local.md` exists, read it after this file.
 
 `agents.local.md` is for personal preferences, student-specific context, and temporary working notes. It should not be committed.
+
+## Data Visualization Guidelines
+- Use color only to encode information, not for decoration.
+- If all bars represent the same type of data, use a single, consistent color.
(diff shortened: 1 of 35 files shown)


annawoodard (approved):
Thank you!!! The purpose of this PR is just to get the code into the repo so Aakrithi can take over, so I'm not doing a full review. Any vestiges we'll clean up at the end of the summer

PR #187, opened 2026-07-16
Contrast-forecasting SSL pretraining - voxel/segment/junction [draft]

spencervenancio:
Self-supervised contrast-forecasting pretext task (adapts [Tang et al. ICLR 2022](https://arxiv.org/abs/2104.08336)) wired for all three graph structures. Stacked on to `feat/implement-new-graphs` it depends on its node-mode refactor, so this PR targets that branch, not `main`.

Validated end-to-end on Duke placeholder skeletons for all three modes. This is plumbing only, not a result (baselines are not strong enough,  and the data is just a placeholder). 33 pretrain + existing GNN tests pass.

Draft: step 2 (model Improvements) and normalization/eval (step 3) still to come.


annawoodard (commented):
@spencervenancio This is very impressive work!!! You came up with a strong idea, and the implementation is thoughtful.

 I especially like that you implemented `PerNodeForecaster` to answer the question: "Does vascular connectivity help forecast contrast, or can each node's own past curve explain its future just as well?" That's exactly the right ablation for testing whether the model is learning something specifically graph-based. I also like the simple persistence and temporal-mean baselines. Together, these give us clear tests of whether the forecasting task is learnable and whether the graph actually contributes.

I found three issues to address before we apply it to the UChicago cohort.

 ### 1. Small masked_mae correctness issue

`masked_mae` says that it accepts a mask that's broadcastable to the prediction shape, but it calculates the denominator before broadcasting the mask. For predictions shaped (nodes, timepoints) and a mask shaped (nodes, 1), the mask is applied across all timepoints in the numerator, while mask.sum() counts each valid node only once. The loss is therefore inflated by the number of timepoints.

If every experiment uses the same forecast horizon and mask shape, this is only a constant scaling factor and probably won't materially affect model rankings or Adam optimization. It could matter when horizons or masks differ, when this loss is combined with another objective, or when its absolute value controls early stopping or gradient clipping. A CPU mask will also fail when the predictions are on a GPU.

The fix should be small: move the mask to the error tensor's device and dtype, explicitly broadcast it to the error_s full shape, and use that broadcasted mask for both the numerator and denominator:

```
  mask = torch.broadcast_to(
      mask.to(device=abs_err.device, dtype=abs_err.dtype),
      abs_err.shape,
  )
```

### 2. Self-supervised pretraining shouldn't require pCR labels

The self-supervised input path currently requires a pCR-label CSV and a label for every case, even though the label is discarded before forecasting training. This prevents us from using the large unlabeled `uc-uf-pretrain` cohort, which is the main reason to use self-supervised learning. Please decouple pretraining graph construction from `data.y` so that clinical labels aren't required for pretraining cases.

### 3. Preserve the UChicago temporal and enhancement contract

#### Physical time is loaded but dropped before training

The existing data loader already reads `ufast_times_seconds.npy` and constructs a valid physical time axis. The regular kinetic features use it to calculate arrival time, slopes, and AUC.

The pretraining path, however, attaches only `data.node_series`. `ForecastGraph` contains `x_seq`, `target `, and  `edge_index`, but no acquisition times. The GRU therefore treats adjacent frames as equally spaced.

  For example:

  Exam A: 0, 3, 6, 9, 12 seconds
  Exam B: 0, 6, 12, 18, 24 seconds

 The model sees both as frames 0, 1, 2, 3, 4, even though their physical enhancement rates differ. It learns change per frame rather than change per second.

This matters in our data. UFAST sampling is generally regular within each sustained acquisition block, but cadence differs substantially between exams, and there are intentional gaps around contrast administration, here are stats for our pCR dataset:

  - In 178/180 valid exams, intervals within the sustained post-transition acquisition block are stable to within 1% of that exam_s median cadence.
  - In 179/180 exams, the interval after the five baseline frames is substantially longer than the regular cadence. Another 106 exams have a second prolonged interval immediately afterward.
  - Median cadence varies from 3.10 to 10.67 seconds between exams, a 3.4-fold difference. The cohort median is 6.47 seconds.
  - Consequently, five consecutive regularly sampled frames can span approximately 12.4_42.7 seconds, depending on the exam.

A model that receives only frame indices treats a 3-second step, a 10-second step, and a 60-second protocol gap as equivalent. It can therefore conflate acquisition protocol with physiology.

Please carry the physical time axis into `ForecastGraph` and provide elapsed time or delta t to the temporal model. If the target times vary, the decoder should also know the times at which it's being asked to predict.

  #### The current window can discard most UFAST frames

`split_forecast_window` always uses the first `input_len + target_len` frames and ignores everything afterward.

With input_len=3 and target_len=2, only five frames contribute, even if the exam contains dozens. More importantly, the first five UChicago frames are all baseline. With the current default, the model would learn to forecast baseline signal rather than contrast arrival.

We should allow the full UFAST curve to contribute. Possible strategies include sliding windows, sampled forecast cutoffs, variable-length prefixes with masks, or windows chosen to span baseline, wash-in, peak, and washout.

I'd start by sampling different valid forecast cutoffs during training. This lets the model use the full sequence and adds useful stochasticity. We should make sure the sampling policy represents the important kinetic portions of the curve rather than generating mostly baseline or plateau examples.

  #### Use the UChicago baseline convention

  The current node-series code subtracts frame 0:

```
  curve - curve[0]
```

  Our UChicago contract instead uses:

  - The mean of the five protocol-defined baseline frames.
  - Relative signal change, (S(t) - S0) / S0.
  - Physical acquisition time in seconds.

The regular kinetic-feature path already follows this contract, but pretraining currently bypasses it.

I wouldn't z-score each curve because that would remove meaningful enhancement amplitude. I also wouldn't normalize each timepoint independently because that would obscure the dynamics we're trying to learn. A single baseline-relative transformation preserves the temporal shape and relative enhancement amplitude while reducing arbitrary scanner intensity scale.

Please have the pretraining path use `baseline_frame_count` and `relative_enhancement` from the same metadata as the regular kinetic-feature path.

  So, the full temporal/enhancement fix is:

  1. Apply the cohort's baseline and relative-enhancement contract.
  2. Carry physical timestamps into the forecasting objects.
  3. Condition the model on elapsed time.
  4. Use windows or masks that allow the full UFAST series to participate.

 After these changes, the next steps toward the full pipeline are batched GPU training over uc-uf-pretrain, checkpoint saving, and a defined method for transferring the pretrained encoder into the pCR model. To measure the benefit of pretraining, we'll compare it against the identical downstream architecture initialized randomly and trained with the same splits and settings.

spencervenancio:
### 1. Small masked_mae correctness issue

Fixed in [eedace0](https://github.com/dsi-clinic/vanguard/pull/187/commits/eedace0de4b10fe0a056f63c0776c8efa9f85ba5). Moved the mask onto the error tensor's device/dtype and `torch.broadcast_to` it to `abs_err.shape` once, then used that single tensor for both numerator and denominator — so a `(N, 1)` mask over `(N, target_len)` predictions no longer inflates the loss by `target_len`, and a CPU mask now applies to a CUDA prediction instead of raising. Added a regression test asserting the `(N, 1)` mask equals its expanded `(N, target_len)` form (scores 6.0, was 12.0 pre-fix) plus a CPU-mask/CUDA-pred test. Good catch, thank you!

### 2. Self-supervised pretraining shouldn't require pCR labels

Done in [f9d6250](https://github.com/dsi-clinic/vanguard/pull/187/commits/f9d62502c734c37fb3983dc98df912ca3425578f). Decoupled forecasting graph construction from `data.y`: `_finalize_data` sets `data.y` only when a label is present, and `_build_case` / `build_case_data` / `_attach_node_features` / `discover_forecast_tasks` now take `label` / `labels_path` as optional (`--labels-path` is optional on the CLI). Label-free cases build a full forecasting graph with no `data.y`; the classification path always passes a label, so it's byte-for-byte unchanged. The one place the label was actually used — the `pcr_dummy` leak-canary node feature — now raises loudly if requested without a label rather than silently. This unblocks `uc-uf-pretrain`.

### Preserve the UChicago temporal and enhancement contract

Implemented across three commits, matching the four-step contract you laid out:

**Baseline / relative-enhancement ([9375fb8](https://github.com/dsi-clinic/vanguard/pull/187/commits/9375fb8254ce0f61ebc60fe4f433ec501e87834d)).** Extracted the curve $\rightarrow$ enhancement transform into `gnn.kinetics.baseline_relative_curve` as the single source of truth; the pretraining node-series now goes through it with the case's `baseline_frame_count` + `relative_enhancement` (mean-of-baseline-frames S0, `(S-S0)/S0`), same as the kinetic path. Segments reference per-voxel then average (nonlinear under relative enhancement, so this matches the kinetic segment path). Legacy `bfc=1`/absolute reduces exactly to the old `curve - curve[0]`, so only real UChicago metadata changes behavior.

**Physical time carried + model conditioned ([e07d4b4](https://github.com/dsi-clinic/vanguard/pull/187/commits/e07d4b4b73647e086b7a300bedf90d27b7fb2965)).** `_build_case` attaches `data.node_times` (the same `time_axis` the kinetics path uses); `ForecastGraph` now carries `input_times`/`target_times`. The encoder appends each input frame's elapsed seconds as a $\Delta t$ channel, and the decoder is time-conditioned — it predicts "enhancement at $+\tau$ seconds" and knows the target query times (which vary per window). A more principled time aware *recurrence* (GRU-D) is noted in #200.

**Full UFAST curve participates ([bca45d5](https://github.com/dsi-clinic/vanguard/pull/187/commits/bca45d5a7168e029f421d6421c52acafaa1a76ad)).** Replaced the single leading window with non-overlapping tiling (similar to Tang et. al.) over the post-baseline region: the bulk of the flat baseline is dropped as windows (kept only for S0, plus ~1 frame as wash-in onset context), each case yields multiple tiles, and times are rebased per tile so tiles carry their real cadence. The train/val split is patient-level (all of a case's tiles on one side) to avoid subject leakage.

Everything here is implementation + unit tests on synthetic data — no real-data validation yet (UChicago pipeline still pending). Horizon is defaulted to a placeholder `2/2` with a horizon-validation sweep recorded as committed follow-up.

annawoodard (approved):
@spencervenancio Thank you! this is great work. You addressed all three parts of my review carefully and thoroughly: `masked_mae` now handles broadcasting and device placement correctly, pretraining no longer requires pCR labels, and the forecasting path now preserves the UChicago baseline-relative enhancement convention, physical acquisition times, and the postcontrast dynamics across the full curve. I especially appreciate that you didn't just patch the immediate issues; you created shared abstractions, time-conditioned the model, and kept each case's windows together during splitting. This was a substantial amount of work, and you handled it thoughtfully. The fixes address my concerns. Great job!

PR #189, opened 2026-07-17
A the dataset adapter in graph extraction's discovery seam (Step 4a in Modularization)

SBose-12:
## Summary
- Part of the multi-dataset modularization (Step 4/6 — see `cohorts/README.md`).
- Adds an optional `adapter: DatasetAdapter | None` parameter to `graph_extraction.pipeline.run_study_pipeline`.
- When an adapter is provided, per-timepoint segmentation discovery routes through `adapter.load_segmented_timepoints()` and MIP-only visualization flipping through `adapter.viz_flip_spec`, instead of the hardcoded MAMA-MIA function/constant.
- When `adapter=None` (every current caller), behavior is byte-for-byte unchanged — this is purely an additive seam, not a behavior change.
- Adds `DatasetAdapter.load_segmented_timepoints` to `cohorts/base.py` (delegates to the existing MAMA-MIA discovery function by default).

## Test plan
- [x] `pytest -q` — 134 passed / 2 skipped (2 pre-existing `torch_geometric` skips, unrelated)
- [x] `ruff check` / `ruff format` clean
- [x] Rebased onto current `main`; verified no behavior change on the `adapter=None` path
EOF
)"

PR #190, opened 2026-07-17
Adopt the dataset adapter in the segmentation stage's seams(Step 4b of Modularization)

SBose-12:
## Summary
- Part of the multi-dataset modularization (Step 4/6 — see `cohorts/README.md`). Stacked on #189 (Step 4a) — review that PR first; this diff only adds the segmentation-stage changes on top.
- Adds an optional `adapter: DatasetAdapter | None` parameter through `segmentation/batch_segmentation.py`'s pipeline (`preprocess_image`, `build_output_path`, `preprocess_parallel`, etc.):
  - Geometry reorientation comes from `adapter.preprocess()` instead of the hardcoded MAMA-MIA axis transform when an adapter is given (UChicago ships already-oriented data and overrides this to a pass-through).
  - The output `source` directory comes from `adapter.case_dataset_name()` instead of the `case_id.split("_")[0]` prefix heuristic.
- When `adapter=None` (every current caller), both paths are byte-for-byte unchanged.
- Includes the **UChicago RAS→LAI orientation fix** in `cohorts/uchicago.py` (`UChicagoDataset.preprocess`): converts SimpleITK's `(z, y, x)` order to the vessel model's `(y, x, z)` without re-applying MAMA-MIA's anatomical flip, since UChicago's manifest images are already reoriented upstream by HFDP. Two earlier attempts (identity pass-through, and a header-derived flip that silently mirrored left/right) are documented in `cohorts/README.md` so they aren't re-derived. Verified quantitatively against BreastDivider ground truth: mean breast Dice 0.902 (correct) vs. 0.174 (mirrored) vs. 0.419 (pass-through).

## Test plan
- [x] `pytest -q` — 134 passed / 2 skipped
- [x] `ruff check` / `ruff format` clean
- [x] Rebased onto current `main`; `adapter=None` path unchanged
- [x] `scripts/validate_uchicago_orientation.py` (job 12894139) — quantitative orientation check vs. ground truth
EOF
)"

### DIFF
--- a/cohorts/README.md
+++ b/cohorts/README.md
@@ -59,20 +59,25 @@ Genuinely different, so it overrides the handful of methods that differ:
 `load_labels()`, and `load_folds()` are all **manifest-driven** (read from a CSV,
 including a `phase_files` list per row); it sets `default_split_policy =
 "provided"` (ships patient-grouped folds) and `report_by = "dataset"`
-(sub-source breakdown). `preprocess()` is a **documented pass-through** — the
-manifest's phase files are already preprocessed upstream (`policy_name =
-hfdp_t1_v1`), so no repo-side transform is applied (and, importantly, the base
-MAMA-MIA orientation transform is *not* used, which would be wrong here). The
-181-exam student manifest lives at
+(sub-source breakdown). `preprocess()` flips array axes 0 and 2 and then applies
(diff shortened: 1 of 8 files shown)


SBose-12:
"Folding into #192  as a single integration PR per Anna's request

PR #188, opened 2026-07-17
Own the full HR→UFAST DICOM preprocessing chain

SBose-12:
# Own the full HR→UFAST DICOM preprocessing chain

## Summary

Vanguard now owns the complete preprocessing path from paired raw DICOM to a vessel skeleton on the UFAST kinetic grid:

`prepare → infer → tc4d → map → qc`

- Loads the exact reviewed native high-resolution (HR) and ultrafast (UFAST) series from DICOM and preserves physical acquisition times and spatial geometry.
- Translation-aligns every HR phase to HR phase 0 before frozen-model preprocessing, then gives all HR vessel-probability phases to TC4D.
- Resamples UFAST to a 1 mm grid and composes accepted translation motion into that single interpolation. UFAST used for kinetics isn't clipped, normalized, or z-scored.
- Maps the static HR skeleton onto the UFAST grid through the shared DICOM `FrameOfReferenceUID` and records image-content alignment QC.
- Fails closed instead of reusing outputs when the policy, case row, manifests, inventory, models, or expected artifacts don't match.
- Requires an explicit `pass` or `manually_approved` alignment status before Vanguard kinetic features can be sampled. The final alignment gate includes HR-to-UFAST alignment, HR interphase motion, and UFAST interphase motion. Legacy cohorts without Vanguard kinetic metadata keep their existing behavior.

## Data contract

The staged Karczmar-lab package contains 179 runnable exams with distinct HR and UFAST series. One shared HR/UFAST acquisition and one split-series exam are retained in the manifests but explicitly excluded from this runnable cohort.

## Validation at this head

- Ruff and all 33 focused preprocessing, DICOM-staging, and GNN data-loader tests pass.
- Real-DICOM v5 prepare smoke: Slurm job `13027311`, reviewed SimBioSys case.
  - Preserved 6/6 HR phases, 13/13 UFAST phases, and all physical UFAST acquisition times.
  - HR-to-UFAST alignment, HR motion, and UFAST motion QC all passed.
  - UFAST motion accepted only bounded, correlation-improving translations; rejected proposals retained the directly resampled raw phase.
- Independent dynamics check: Slurm job `13028748`.
  - Raw DICOM versus direct 1 mm resampling relative mean-curve correlation: 0.99999993.
  - Direct resampling versus final motion-corrected relative mean-curve correlation: 0.99999967.
  - Peak enhancement amplitude ratio: 1.0; peak phase remained 10.
  - Maximum relative curve difference from accepted motion correction: 0.000683.

[...]

PR #192, opened 2026-07-17
Adapter seams across the DICOM-to-skeleton-and-kinetics pipeline (graph extraction, segmentation, QC/CLI) (Step 4 of Modularization)

SBose-12:
## Summary
Combines Step 4 of the multi-dataset modularization (see `cohorts/README.md`) into one integration PR, at Anna's request. This was originally three stacked PRs (#189 graph extraction, #190 segmentation, #191 QC/CLI) — that split was reverted because these three seams collectively define a single end-to-end DICOM-to-skeleton-and-kinetics pipeline: they touch the same discovery/orientation/reporting contract in `cohorts/base.py`, and merging them separately would leave the pipeline in an inconsistent state between merges and require repeated rebasing of the later stages on the earlier ones. This is an intentional, one-off exception to our normal preference for small atomic PRs; we're returning to that afterward.

- **Graph extraction** (formerly #189): optional `adapter` param on `run_study_pipeline`; per-timepoint discovery and MIP-flip routed through the adapter when given.
- **Segmentation** (formerly #190): optional `adapter` param through `batch_segmentation.py`'s preprocessing/output-path functions; includes the UChicago RAS→LAI orientation fix (verified against BreastDivider ground truth, mean Dice 0.902).
- **QC/CLI** (formerly #191): adapter-driven QC `report_by` seam, broadened `split_col` collision check, CLI wiring for adapter selection, Slurm smoke-test scripts.
- All three stages preserve byte-for-byte MAMA-MIA behavior when `adapter=None` (every current caller).

## Test plan
- [x] `pytest -q` — 141 passed / 2 skipped
- [x] `ruff check` / `ruff format` clean
- [x] Rebased onto current `main`; `adapter=None` path unchanged
- [x] Slurm smoke test on 5 UChicago patients through segmentation + skeleton stages
EOF
)"


annawoodard (changes requested):
You've done a tremendous amount of careful work here, Sarit!!! I'm especially impressed that you caught the cross-patient intermediate-file collision, fixed patient limiting at the case level, strengthened the fold validation, and tested that the adapter-free path preserves existing MAMA-MIA behavior. The smoke testing also did exactly what we wanted: it exposed real orientation and memory problems before a larger run. This is strong, thoughtful engineering.

There are a few changes we need before merging, mostly because I merged PR #188 after you built this branch, so the canonical UChicago pipeline has changed underneath you.

First, please rebase onto current main. There's a conflict in `batch_segmentation.py`, and we need to preserve both sides of it: your adapter should determine the spatial reorientation, and then the result should go through PR 188's shared `frozen_model_intensity_preprocess`. We don't want to restore the older duplicated normalization code.

The more important change is that we should no longer use the legacy UChicago UFAST NIfTI files as inputs to vessel segmentation and skeletonization. PR #188 established the pipeline we now want:

  - Segment every higher-spatial-resolution HR phase.
  - Construct the skeleton from the HR vessel probabilities.
  - Map that static skeleton onto the motion-corrected UFAST grid.
  - Compute kinetics from the raw UFAST signal using the physical timestamps.

The UChicago path in this PR currently segments the UFAST phases directly, then uses vessel probabilities and filename indices for kinetics. That was a reasonable implementation when you started this work, but it's been superseded by the raw- DICOM HR_UFAST pipeline. Please keep the generic adapter, collision-prevention, CLI, and QC work, but remove or disable the legacy UChicago imaging route. Ideally, selecting UChicago here should direct the user to the paired raw-DICOM preprocessing pipeline so we don't leave two competing implementations.

One smaller correction: if an adapter requests report_by="dataset" but that column is missing, the evaluation should raise a clear error instead of silently falling back to another subgroup column. Otherwise, we could think we generated the requested UChicago sub-source QC when we didn't.

I tested a combined version using the correct conflict resolution: Ruff passed, 81 focused tests passed, and all nine standalone segmentation seam checks passed. So the underlying adapter work is in very good shape. Most of what you built should stay. We mainly need to rebase it around the newer UChicago preprocessing contract.

This was a large and difficult integration, and you've handled many subtle issues well. The collision fix in particular prevented a serious silent data-corruption bug. Great work!!!

annawoodard (approved):
THIS IS GREAT WORK!!!! you're the best @SBose-12 

👏 👏 👏 👏

PR #193, opened 2026-07-20
 Require the dataset adapter everywhere, retire None fallbacks (Step 5/5 of Modularization)

SBose-12:
## Summary

Step 5 of the multi-dataset migration: every pipeline stage that adopted the
adapter in Steps 2-4 (`tabular/cohort.py`, `tabular/train.py`,
`modeling/ablation.py`, `segmentation/batch_segmentation.py`,
`graph_extraction/pipeline.py`) now *requires* one instead of tolerating
`adapter=None` — the hardcoded MAMA-MIA fallback logic those functions used
to fall back to is deleted, not just made unreachable.
`cohorts/factory.py` gains `require_adapter_from_config` /
`require_imaging_adapter_from_config`, which raise a clear error naming the
missing `dataset:` block instead of returning `None`.

Two real bugs surfaced by making the adapter mandatory, both fixed here
(neither was live in production before this branch, but both would have
broken it the moment the adapter became required):

- `MamaMiaDataset` had no "all cohorts" mode. Production imaging Slurm jobs
  process the whole MAMA-MIA tree across all four cohorts in one job; forcing
  a single cohort choice would have silently scoped those jobs to 1/4 of the
  data. `cohort=None` now means no discovery filter, verified byte-identical
  to the old unfiltered `find_nii_files` walk.
- `_base_name_for` in `batch_segmentation.py` unconditionally prefixed the
  intermediate filename with `case_id` whenever an adapter was given —
  correct for UChicago's collision-prone flat filenames, but wrong for
  MAMA-MIA (whose filenames already embed the case id): output would have
  come out `DUKE_001_DUKE_001_0000` instead of `DUKE_001_0000`. Fixed with a
  regression test.

Also closes a Step 4 follow-up: `modeling/ablation.py` built an adapter but
never applied provided folds / patient grouping / `report_by`, so an
ablation matrix over a dataset with provided folds silently ignored them.
Brought to parity with `tabular/train.py`, with tests pinning why folds are
applied once per arm but group keys resolve fresh per (arm, family, mode).

24 MAMA-MIA tabular configs gained a `dataset:` block; three imaging Slurm
scripts now pass `--dataset-name`/`--dataset-root` (with an unrelated
pre-existing `--study-id`/`--case-id` bug fixed in passing); the now-obsolete
`scripts/validate_adapter_feature_parity.py` (its whole purpose was comparing
`adapter=None` vs. `adapter=<real>` output) is deleted.

**Base note:** targets `feat/adapter-in-qc` (#192), not `main` — #192 hasn't
merged yet and this branch depends on its Step 4 work (the imaging-route
retirement, `report_by` fail-loud behavior, etc.). Retarget to `main` once
#192 merges.

**Open with Anna** (carried over, not resolved here): whether `cohort`
selection belongs somewhere other than the adapter for multi-cohort tabular
runs — see `modularization-status.md` item 5.

## Test plan

- [x] `pytest tests/ -q` — 173 passed / 2 skipped (up from 165/2; the 2 skips
      are pre-existing `torch_geometric` importorskips)
- [x] `segmentation/tests/test_adapter_seam.py` — all 9 checks pass,
      including the new double-prefix regression check
- [x] `segmentation/tests/test_preprocess_parallel.py` — passes
- [x] `ruff check` / `ruff format --check` clean on every changed file
- [ ] Real Slurm smoke run of the updated imaging scripts (only exercised by
      CPU-only synthetic-data tests so far, not an actual GPU job)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


annawoodard (approved):
Great work, Sarit!! You've followed the adapter contract through segmentation, graph extraction, tabular modeling, ablations, configs, and Slurm, and I especially like that the UChicago imaging route now fails closed instead of allowing the wrong pipeline to run silently. I found just two concrete things to fix. First, `slurm/submit_improved_feature_selection.slurm` still calls `prepare_data(config, outdir)` without the now-required adapter, so that job will fail with a missing-argument error. Please build the adapter with `require_adapter_from_config(config)` and pass `adapter=adapter`. Second, please update the graph-extraction commands in the root `README` and `graph_extraction/README.md`: they now need `--dataset-name` and `--dataset-root`, and the graph README still uses `--study-id` rather than `--case-id`. Finally, #192 has now been squash-merged, so please merge or rebase current main into this branch and resolve the resulting conflicts. Once those are addressed and CI passes again, please merge. I'm approving, assuming you'll address these before merging. Great job!!

PR #194, opened 2026-07-21
DUKE Symmetry Lines v2: Updated to Remove Signal from Technical Confounders

aakrithiram:
## Summary

  Adds `asymmetry_predictions_v.2`, the nuisance-filtered DUKE vessel asymmetry pCR model.

  This version starts from the original vessel asymmetry feature set (`asymmetry_predictions_v.1`, previously
  referred to as `logreg_all`) and removes features with high correlation to nuisance variables.

  ## What Changed

  - Added `asymmetry_predictions_v.2/`
  - Added a reproducible build/evaluation script:
    - `asymmetry_predictions_v.2/run_asymmetry_predictions_v2.py`
  - Added filtered model input:
    - `asymmetry_predictions_v.2/asymmetry_predictions_v2_features.csv`
  - Added model outputs:
    - `cv_metrics.csv`
    - `oof_predictions.csv`
    - `run_metadata.json`
  - Added nuisance audit files:
    - `removed_nuisance_features.csv`
    - `all_feature_nuisance_scores.csv`
    - `filtered_feature_nuisance_scores.csv`
  - Added plots comparing v1 and v2 nuisance sensitivity.

  ## Feature Filtering Rule

  For each vessel/asymmetry feature, compute the maximum absolute Spearman correlation to:

  - `xy_spacing_mm`
  - `z_spacing_mm`
  - `tumor_size_tumor_voxels`

  Features with max nuisance `|r| >= 0.25` are removed.

  ## Results

  Original model, `asymmetry_predictions_v.1`:

  - Features: 115
  - Mean AUC: 0.628 ± 0.042
  - Max nuisance |r|: 0.384

  Updated model, `asymmetry_predictions_v.2`:

  - Features: 104
  - Removed features: 11
  - Mean AUC: 0.597 ± 0.044
  - Mean AP: 0.311 ± 0.016
  - Max nuisance |r| after filtering: 0.228

  ## Interpretation

  `asymmetry_predictions_v.2` trades some within-DUKE AUC for lower nuisance sensitivity. The goal is to improve
  generalizability by reducing dependence on spacing, tumor size, and other non-biological effects.

  ## Notes

  No raw imaging data are modified. The added CSVs and plots are derived tabular artifacts.


annawoodard (changes requested):
Aakrithi, this is really great work!! 👏 👏 👏 

You've made the filtering rule explicit, kept imputation and scaling inside the CV folds, saved the OOF predictions and audit tables, and clearly reported the AUC tradeoff instead of overselling it. I only have two substantive requests.

1. First, your current analysis establishes that each retained feature has marginal Spearman (|r| < 0.25) with the audited variables, but that isn't quite the same as showing that the multivariate model itself has lower technical-confounder sensitivity or will generalize better outside Duke. Multiple weakly associated features can still jointly encode spacing. Please either narrow the wording to "lower maximum marginal feature_confounder correlation" or add a small OOF model-level audit relating `y_prob` to `xy_spacing_mm` and `z_spacing_mm`. I'd describe tumor size separately as a biological/clinical covariate included to test whether the vessel signal is independent of tumor extent, rather than calling it a technical confounder. (That distinction is important so let's discuss on Slack if you have questions).
2. Second, you don't need to reproduce Sarit's full upstream pipeline or commit the input tables, but please add a short README section documenting the two derived inputs: what they contain, which workflow produced them, their shared paths and checksums, and the exact command you ran. If the feature table currently exists only in an untracked checkout, please copy that small CSV to a durable shared location. The script also doesn't appear to generate the separately listed v1/v2 nuisance heatmaps; you can simply identify them as separately generated exploratory artifacts rather than expanding the script.

Besides that, everything looks great!

annawoodard (approved):
Aakrithi, thank you! This fully addresses my feedback. You narrowed the interpretation appropriately, clearly separated the technical confounders from tumor size as a biological/clinical covariate, and added the OOF model-level audit I requested. You also documented the derived inputs, provenance, durable shared paths, checksums, and exact command, and clarified the status of the exploratory heatmaps. I verified that the shared files and checksums are correct. This is careful, thoughtful work. Great job!

PR #195, opened 2026-07-22
Run the GNN on UChicago ultrafast data

spencervenancio:
Wires the GNN pipeline to the UChicago internal ultrafast DCE dataset via the
merged dataset adapter (`cohorts/`, #193).

### What's here (2 commits)
- **Ultrafast kinetic contract** in the data loader: `gnn_kinetic_baseline_frame_count`
  + `gnn_kinetic_relative_enhancement` make every case use relative signal change
  `(S-S0)/S0` over a fixed precontrast baseline, overriding per-case `run_summary.json`.
  `None` (default) preserves the existing per-case MAMA-MIA behavior — fully backward compatible.
- **Build submit script** updated to pass the ultrafast contract through.

### Validation
- Ran the 2-feature GNN end-to-end on UChicago ultrafast via Sarit's `preprocessing_out/centerlines`.
- `pytest tests/test_gnn_data_loader.py` — covers the kinetic contract.

### Notes
- Default behavior on MAMA-MIA is unchanged (new keys default to the legacy path).
- Overlaps `config.py` with the tier PR (both add `DEFAULT_CONFIG` keys); trivial keep-both resolution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,3 +235,6 @@ presentations/
 
 # Design Docs
 docs/design/
+
+# Dev-phase decision gates (throwaway scaffolding, deleted at end of dev)
+gnn/dev_gates/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,6 +80,8 @@ Derived outputs should go in clearly named output folders with enough provenance
 - Prefer saved CSVs, plots, QC panels, and short README/provenance notes for important outputs.
 - Document split/CV policy when training or evaluating models.
 - Do not change cohort definitions, labels, or train/test splits without making the change explicit.
+- Do not commit evolving or personal working docs — migration plans, scratch notes, study guides, TODO/roadmap files. These clutter branches and PRs and go stale. Keep them in a gitignored location (`docs/design/`, `agents.local.md`, or an untracked scratch path), not in `docs/` or the repo root.
+- Committed docs must be durable references others rely on: READMEs, data-layout/policy docs, provenance/results write-ups (e.g. `docs/issue*.md`), and **decision records that code or configs point to** (e.g. `gnn/DESIGN_segment_graph.md` and `gnn/PLAN_advanced_modeling.md`, which `config.py` cites by name). A "plan" or "design" file earns a committed spot only as a stable decision record referenced from code — not as a work-in-progress roadmap.
 
 ## Local Preferences
 If `agents.local.md` exists, read it after this file.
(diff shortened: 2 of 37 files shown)


annawoodard (approved):
Spencer, this is great work!! You've connected the UChicago data to the GNN with the correct relative-enhancement definition, preserved the manifest's patient-grouped folds, made the cache contract explicit, and added a useful labels exporter and focused tests. One non-blocking warning for the current run: Sarit's existing preprocessing-out summaries predate the complete Vanguard alignment/provenance contract, so they don't contain alignment_qc_status or kinetic_feature_policy; using the override is reasonable for now given the project timeline, but we should treat these results as preliminary with respect to HR-to-UFAST alignment. You also shouldn't ultimately need to know or manually configure the baseline-frame count: it's acquisition metadata, not a model hyperparameter. The reviewed UChicago paired-case manifest already records `ufast_baseline_frame_count`, and Sarit's updated Vanguard preprocessing should write that value, the enhancement policy, and alignment approval into each case's `run_summary.json` for the GNN to read automatically. Once Sarit's outputs contain that information, please remove the temporary kinetic override and its duplicated YAML/CLI/Slurm settings rather than maintaining compatibility with the legacy output format we won't use. Anytime we know the answer, we always want to avoid asking the user for it, because users make mistakes 😄. Longer term, the cache builder could read the same YAML as training and reserve CLI arguments for execution controls such as worker count and force-rebuild, but that's a design note rather than a request to refactor it during your remaining time. Overall, this looks great and I think we're ready to go as-is, you can implement that feedback if you have time but none of it is blocking.

spencervenancio:
> Sarit's existing preprocessing-out summaries predate the complete Vanguard alignment/provenance contract, so they don't contain alignment_qc_status or kinetic_feature_policy;

I ran @SBose-12's updated skeletonization, so now outputs contain `alignment_qc_status` and `kinetic_feature_policy`. 

>  Once Sarit's outputs contain that information, please remove the temporary kinetic override and its duplicated YAML/CLI/Slurm settings rather than maintaining compatibility with the legacy output format we won't use.

Completed in [6f3f6ce](https://github.com/dsi-clinic/vanguard/pull/195/commits/6f3f6cea2be6a2a340240d1d82069e5532363086).

> Longer term, the cache builder could read the same YAML as training and reserve CLI arguments for execution controls such as worker count and force-rebuild, 

This is deferred to issue #198.

PR #197, opened 2026-07-22
Advanced GNN modeling Tier 0,1: training hygiene + multi-statistic readout

spencervenancio:
This PR adds complexity the baseline GNN model. It's part of a 3 tiered plan laid out in `gnn/PLAN_advanced_modeling.md`.  I will roll out tiers 2 and 3 in a future PR.  

The additions are summarized as follows: 

**Training hygiene.** The GNN's training loop was bare-bones: plain Adam, plain BCE loss, a fixed number of epochs, and it reported the model at its final epoch even if that epoch was already overfit. This adds the standard training controls — class-weighted loss (pCR-positive is the minority class), weight decay, gradient clipping, a learning-rate scheduler, early stopping, and picking the best epoch by validation loss instead of taking the last one. Each control defaults to the old behavior, so existing runs are unaffected unless a config turns it on.

**Better graph readout.** Before the classifier, the model averaged every node down to a single mean, which discards the peaks and spread that carry the signal — a plain tabular model on those same statistics was outperforming the GNN. This lets the readout combine the mean with max and sum pooling, so the graph keeps its extremes, not just its average. The default stays mean-only, so it's backward compatible.

**How it runs.** New config files define the experiment arms (a frozen baseline reproducing the old behavior, a training-hygiene arm, and a readout arm) plus the Slurm scripts to launch them. Both new code paths have unit tests.

**One honest result.** A single-cohort (DUKE) re-run showed the readout gain did not hold up — it was a cross-cohort artifact, not a genuine improvement. That's logged in the project lab notebook. It does not block this PR: the readout is off by default, and this is exactly why it stays gated behind the frozen-baseline comparison rather than being switched on.

**Backward compatibility.** Every new option defaults to the historical behavior. The shared config defaults did shift, so the baseline config files pin the old values explicitly to keep reproducing the original numbers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,3 +235,6 @@ presentations/
 
 # Design Docs
 docs/design/
+
+# Dev-phase decision gates (throwaway scaffolding, deleted at end of dev)
+gnn/dev_gates/
--- a/config.py
+++ b/config.py
@@ -162,6 +162,14 @@ DEFAULT_CONFIG: dict[str, Any] = {
         # gnn/DESIGN_segment_graph.md.
         "gnn_node_mode": "voxel",
         "gnn_node_features": ["peak_time", "radius"],
+        # Whole-graph readout for GCNClassifier (voxel/segment): "mean"
+        # (default, the original single global_mean_pool -- fully backward
+        # compatible), "mean_max", or "mean_max_sum". Concatenating max/sum
+        # pools keeps the distributional tails that mean-pool discards; see
+        # gnn/PLAN_advanced_modeling.md decision D2.1 and gnn/model.py
+        # POOLING_WIDTHS. Junction mode (EdgeGNNClassifier) ignores this and
+        # always mean-pools for now.
+        "gnn_pooling": "mean",
         # Edge features for node_mode="junction" (segment-as-edge, Option A):
         # the segment summary rides on the edges there. Empty for voxel/segment
         # mode, which carry no edge features. Uses the seg_* vocabulary (see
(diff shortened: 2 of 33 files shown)


annawoodard (approved):
@spencervenancio This is excellent work! You've thought carefully about isolating each modeling change, pairing folds and seeds, preserving a frozen baseline for comparison, and reporting the Duke null result honestly. The configurable readout is also implemented cleanly and tested well. I found one issue we should fix before merging: the outer held-out fold is currently used for the plateau scheduler, early stopping, and selection of the best epoch, and then that same fold supplies the reported OOF predictions and AUC. That creates model-selection leakage because the held-out labels influence which checkpoint is evaluated, so the reported CV performance can be optimistically biased. Please make an inner training/early-stopping split within each outer training fold and keep the outer fold completely untouched until the final prediction. We don't need to preserve the behavior of every older GNN config as we move forward, but the PR shouldn't say all existing runs are unaffected: gnn.yaml can remain explicitly frozen for the scientific comparison while the other configs adopt the improved shared defaults. One quick fail-fast improvement would also be to reject an unrecognized `lr_scheduler` value instead of silently treating it as none. Finally, I'd soften "it was a cross-cohort artifact" to "the apparent full-cohort gain didn't replicate when retrained on Duke alone." The Duke result establishes non-replication, but it doesn't by itself identify cohort mixing as the cause. 

Great work! To keep things moving, I'm going ahead and approving, I assume you'll address these points and then merge

PR #199, opened 2026-07-24
Agents/comment style

spencervenancio:
I noticed that my agent was using comments too much like agentic memory. This PR adds a line to `AGENTS.md` and cleans up some specific comments, restoring them to the proposed convention.


annawoodard (commented):
omgggg love this i'm adding it to my agents. every time i read something like that i'm annoyed but hadn't gotten around to fixing it yet. thanks spencer 😄

annawoodard (approved):

PR #201, opened 2026-07-27
Merge UFAST-direct vessels into the HR skeleton as the pipeline's UChicago output

SBose-12:
## Summary
- Runs the frozen vessel model + TC4D over UFAST phases directly (not just via HR),
  so vessels the HR route misses can be recovered from UFAST's own signal.
- Merges the two TC4D skeletons with an HR-authoritative policy: every HR-mapped
  voxel is always kept; a UFAST-direct voxel only adds new coverage if it isn't
  already near an HR voxel (proximity match, not exact-voxel equality — see
  `preprocessing/merge.py` module docstring) and it clears a component-size or
  contrast-kinetics filter. This is the mechanism that prevents the merge from
  double-drawing the same vessel when both routes pick it up slightly offset.
- Promotes this from an `analysis/` exploration script into `preprocessing/pipeline.py`
  as first-class `infer`/`tc4d`/`map` stages — the merged skeleton is now the
  pipeline's default centerlines output, with the pure HR-only skeleton preserved
  alongside (`*_skeleton_4d_exam_mask_hr_only.npy`) so nothing is silently discarded.
- Fixes a bug in the earlier exploratory merge where HR-only voxels could be
  silently dropped by TC4D's own cleanup pass, contradicting the "HR is
  authoritative" policy.
- Wires up `write_temporal_mip` (a skeleton-free UFAST MIP, previously dead code)
  into `qc_case` and documents the merge step, required UFAST route, and new
  per-case output files in `preprocessing/README.md`.


annawoodard (approved):
Sarit, this is really strong final integration work. You’ve taken the exploratory HR/UFAST work and turned it into a documented production pipeline while preserving the HR-only result and recording the source of the merged voxels. The HR-authoritative, UFAST-additive policy is clear, and the temporal MIP is a useful QC addition. I only found two substantive issues before merging.

First, please update the production cohort runner to match the new two-route contract. infer_case now records hr_phases and ufast_phases, but preprocessing/run_cohort.py still reads inference["phases"], so it’ll fail when checking a completed case for resubmission. Its TC4D, mapping, and QC checks also recognize only the old HR-only artifacts, so they can incorrectly skip a case that’s missing the UFAST TC4D result, merge provenance, preserved HR-only output, or temporal MIP. Please also bump the preprocessing policy from v5 to v6 so existing HR-only v5 outputs can’t be silently reused. Finally, the actual cohort Slurm scripts under slurm/run_paired_preprocessing_* still have four-hour GPU and eight-hour CPU limits. Those are the production entry points and should receive the increased limits, especially since UFAST TC4D alone is documented as needing about 11.5 hours. 

Second, the confirmed_both provenance label currently uses merged_skeleton & confirmed_ufast_mask. Because proximity-confirmed UFAST voxels are intentionally left out of the merged mask in favor of the authoritative HR coordinates, this labels mostly exact overlaps rather than the documented two-voxel proximity agreement. Please label the HR voxels that are near the UFAST mask instead. This doesn’t change the merged result; it just makes the provenance counts and QC colors accurately represent the merge policy.

This is an impressive amount of careful work, especially the transition from exploratory analyses to a reproducible pipeline!

To keep things moving I'm approving rather than requiring another review iteration, please merge as soon as the above two items are addressed

PR #202, opened 2026-07-27
Matlab translated skeleton addition

SBose-12:
## Summary
- Adds a new `complement` pipeline stage that runs a third, independent vessel
  source — a Python port of the lab's MATLAB SegVessel/Jerman vesselness
  pipeline (`segmentation/matlab_vessel_segmentation`, ported from
  `matlab-conv-2/vessel_pipeline`) — on top of the HR+UFAST merged skeleton.
- Additive-only, same "never override" contract as `preprocessing/merge.py`: a
  MATLAB-route voxel already near the merged skeleton is "confirmed" and
  calibrates a quality gate (enhancement + Jerman vesselness percentile vs.
  the confirmed voxels' own values); a MATLAB-route voxel far from the merged
  skeleton is only added if it clears that gate, sits interior to the breast
  mask, and belongs to a large-enough connected component. Validated in
  `matlab-conv-2/vessel_pipeline/complement_filter.py` across 6 UChicago exams
  (97-640 added voxels/exam, ~0% within 2 voxels of the breast-mask boundary).
- Fixes a geometry bug found in review: `SegVessel`'s internal HR→UFAST-grid
  step (`imresize3_nearest`) is a shape-based nearest resize that assumes
  identical in-plane pixel spacing between the HR and UFAST grids — false in
  this pipeline, since UFAST is resampled onto a 1-mm isotropic grid while
  native HR is finer. `preprocessing/complement.py`'s `_load_hr_subtraction`
  now resamples the HR subtraction volume onto the UFAST grid first, using the
  same `DicomGeometry`-driven physical mapping every other HR↔UFAST step in
  this codebase already uses, so `SegVessel`'s internal resize becomes a
  same-shape no-op instead of a silent misalignment.
- `<exam_id>_skeleton_4d_exam_mask.npy` (the file every downstream consumer
  reads) becomes the complemented skeleton once `complement` runs; the
  pre-complement HR+UFAST merge is preserved alongside as
  `<exam_id>_skeleton_4d_exam_mask_hr_ufast_merge_only.npy`, and per-case stats
  land in `complement_provenance.json` — same "never silently discard"
  pattern the merge step already uses.
- Documents the new stage, its CLI usage, and its output files in
  `preprocessing/README.md`.


annawoodard (approved):
Sarit, this is extremely impressive work! You didn’t just translate the MATLAB vessel pipeline: you integrated it into Vanguard, found and fixed the HR-to-UFAST geometry problem, designed an additive quality gate, and repeatedly challenged the result with controls that could actually falsify it. I’m especially impressed that you caught the contaminated percentile sweep, reran it against the correct pre-complement baseline, tested both real and synthetic motion, and documented where the method still finds ambiguous boundary-like structures. The way you followed unexpected results instead of defending the first implementation is exactly what good research looks like.

I found one substantive downstream issue to fix. complement_case updates the canonical centerline mask but leaves the corresponding support mask unchanged. I checked eight completed cases, and 48-75% of the newly added complement voxels fall outside the saved support mask. The graph pipeline can repair containment by adding those centerline voxels to the support mask, but radius and caliber features will then treat the additions as essentially one-voxel-wide rather than using the vessel width estimated by SegVessel. Please propagate support for the accepted complement branches as well. SegVessel already returns vmask, so I’d restrict that vessel region to the accepted branches and union it into the existing support mask when writing the complemented result. This doesn’t undermine the complement selection or topology results; it makes the saved representation internally consistent and prevents misleading caliber features.

We also need to acknowledge the intellectual and methodological provenance clearly. This implementation extends Zhen’s code, which itself builds on the Wu et al. paper she emailed us. Please cite both Zhen and the Wu et al. paper in the module and function docstrings around the translated MATLAB implementation, and add both citations to the top-level repository README. The current references to the lab’s MATLAB pipeline and the standalone validation workspace aren’t sufficient attribution for the underlying method.

Once those two issues are addressed, please merge. This is an incredible amount of thoughtful and scientifically careful work!!

PR #207, opened 2026-08-01
docs: record UChicago ultrafast cohort paths and caveats in AGENTS.md

spencervenancio:
Document the two published cohort roots (longitudinal 240/196, pretreatment 137/137), the shared `dce2d_internal_ultrafast_manifest.csv` contract, and the per-root `images/centerlines/tumor` layout under Vanguard Data Layout.

Also record the tumor-mask caveats (14 empty longitudinal masks, 69/36 exams flagged as possible bilateral cases needing review, four pending_cases.csv rows) and the restricted raw DICOM sources, which are not deidentified and must stay within Karczmar-lab shares.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -147,6 +147,7 @@ require Huo-lab access.
 
 ## Project Conventions
 - Keep scripts and outputs named clearly enough that another student can understand what they are for.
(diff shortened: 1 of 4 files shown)


annawoodard (changes requested):
This is great @spencervenancio, thank you!!

We want this repo to be usable out-of-the-box by other groups so can we tweak this by moving the data info into a linked uchicago-specific markdown file. That way agents trying to reproduce this externally know they need to either prepare their own dataset or get access to ours etc

We don't have to let perfect be the enemy of the good here by perfectly documenting all data provenance I think exactly what you have is great just make it explicit this is only valid at uchicago and put it in a linked site-specific file

spencervenancio:
I added `UCHICAGO.md` and it is now linked to by `README.md` and `AGENTS.md`.

